Issue #7, opened 2019-07-12
Parallelize matrix calculation


ivalaginja:
The PASTIS matrix calculations take a lot of time and the individual steps are independent of each other, so this offers itself ideally for some speeding up with multiprocessing or -threading.

ivalaginja:
Done with #24

PR #4, opened 2019-07-12
Refactor for final SegmentedDeformableMirror in HCIPy

ivalaginja:
I developed all important PASTIS codes on an old commit of `hcipy` on the `segmented_mirror` branch and want to refactor all that with this PR to work with the final `SegmentedDeformableMirror` class that now got merged into master. I started retiring the old codes and notebooks and making new ones that use the new version.

The problem is that I cannot make the APLC propagation work properly (see [this notebook](https://github.com/spacetelescope/PASTIS/blob/feature/sm_refactor/Jupyter%20Notebooks/LUVOIR/3_Testing%20the%20E2E%20simulations.ipynb) for the LUVOIR example), I get too much light in the dark hole of the LUVOIR APLC. Since this is quite a rabbit hole to dig in to, I am just making this a draft PR that I *might* come back to at a later time (I should, but time flies...).

Note how this means that the environment in which the PASTIS codes are currently run needs to have a very specific commit of `hcipy` installed, which is commit `980f39c`. This PR should bring all the PASTIS codes up to speed with the usage of the most up-to-date version of `hcipy`.

My personal notes for this start on 07/09/2019:
```
Since I needed to develop a custom SM for the ULTRA work, my scripts and notebooks
all kinda use a different version. I think I will retire the entirety of them in a
subfolder and restructure all the rest from scratch. This will take quite a
 while I am afraid but I cannot risk code confusion.
By now, Emiel has merged the segmented_mirror branch into master and updated its usage,
so that will be how I restructure my codes.

At some point I should also make SegmentedTelescopeAPLC() inherit directly from
hc.SegmentedDeformableMirror().
All changes I am making are happening on branch sm_refactor.

PyCharm:

from hcipy.optics.segmented_mirror import SegmentedMirror
… on my outdated segmented_mirror branch on commit 980f39c
 	⁃	luvoir_imaging.py —> new version
from luvoir_imaging import SegmentedTelescopeAPLC
 	⁃	contrast_calc_hicat.py
 	⁃	luvoir_matrix_building_numerical.py
import atlast_imaging as atim
 	⁃	segmirr_test.py   —> moved to depr dev

Jupyter notebooks:

ATLAST and HCIPy:
from hcipy.optics.segmented_mirror import SegmentedMirror
 	⁃	3_Testing the E2E simulations   —> moved to depr dev —> new version
import atlast_imaging as atim
 	⁃	3a_Segmented Mirror for HCIPy   —> moved to depr dev
[...]

PR #22, opened 2019-11-25
Use multiprocessing for PASTIS matrix calculation

ivalaginja:
Adresses #7 

I have adapted this from `hicat.scripts.stroke_minimization.calculate_jacobian`. Works fine, but want to test on full matrix before merging.

I also used this PR to clean up some configfile reading and directory creation.

### DIFF
--- a/pastis/hockeystick_contrast_curve.py
+++ b/pastis/hockeystick_contrast_curve.py
@@ -15,6 +15,8 @@ import time
 import numpy as np
 import pandas as pd
 import astropy.units as u
+import matplotlib
+matplotlib.use('Qt5Agg')
 import matplotlib.pyplot as plt
 
 from config import CONFIG_INI
@@ -49,11 +51,8 @@ def hockeystick_jwst(range_points=3, no_realizations=3, matrix_mode='analytical'
 
     # Set up path for results
     outDir = os.path.join(CONFIG_INI.get('local', 'local_data_path'), WORKDIRECTORY, 'results')
-    if not os.path.isdir(outDir):
-        os.mkdir(outDir)
-
-    if not os.path.isdir(os.path.join(outDir, 'dh_images_'+matrix_mode)):
-        os.mkdir(os.path.join(outDir, 'dh_images_'+matrix_mode))
+    os.makedirs(outDir, exist_ok=True)
+    os.makedirs(os.path.join(outDir, 'dh_images_'+matrix_mode), exist_ok=True)
 
     # Loop over different RMS values and calculate contrast with PASTIS and E2E simulation
     e2e_contrasts = []        # contrasts from E2E sim
@@ -218,6 +217,9 @@ def hockeystick_luvoir(apodizer_choice, matrixdir, resultdir='', range_points=3,
     rms_range = np.logspace(-4, 4, range_points)      # Create range of RMS values to test
     realiz = no_realizations                             # how many random realizations per RMS values to do
     ##########################
+
+    # Create results directory if it doesn't exist yet
+    os.makedirs(resultdir, exist_ok=True)
 
     # Loop over different RMS values and calculate contrast with MATRIX PASTIS and E2E simulation
     e2e_contrasts = []        # contrasts from E2E sim
--- a/pastis/hockeystick_contrast_curve.py
+++ b/pastis/hockeystick_contrast_curve.py
@@ -15,8 +15,6 @@ import time
 import numpy as np
 import pandas as pd
 import astropy.units as u
-import matplotlib
-matplotlib.use('Qt5Agg')
 import matplotlib.pyplot as plt
 
 from config import CONFIG_INI
--- a/pastis/matrix_building_numerical.py
+++ b/pastis/matrix_building_numerical.py
@@ -10,7 +10,10 @@ This module contains functions that construct the matrix M for PASTIS *NUMERICAL
 
 import os
 import time
+import functools
 import numpy as np
+import matplotlib
+matplotlib.use('Qt5Agg')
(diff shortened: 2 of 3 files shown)


ivalaginja:
Hm, it worked fine on 5 segments, but gives me an error after the very last pair when using 120 segments:
Screen Shot 2019-11-25 at 19 44 45
[...]
Screen Shot 2019-11-25 at 19 42 50

It looks like it can't return the result from the pool. Not sure if it is too large or what is happening.

mperrin:
The error text last line says that it’s trying to format something as an integer, specifically a 32 byte integer, but the value is outside the range supported by 32 bit integers.

ivalaginja:
Yes, it seems like `multiprocess.pool` wants to do it as a 32bit integer, but my data is too large:
https://github.com/joblib/joblib/issues/818#issuecomment-445865581

In the link, they suggest using threads instead of processed. Alternatively, I could cut my data. I will have to look into this more thoroughly - if you have any ideas @mperrin, please let me know.

mperrin:
I don’t think I understand yet what your code is doing well enough to comment without taking a closer look... Is that expected that it’s trying to return >2GB objects from each process? Or is it the set of all of them together that is >2GB? 

I have no experience using the threading backend they suggest. I don’t know if that would work well or not, but you could try it. 

Alternatively perhaps restructure your code so the returned values are not passed in memory but instead are written as temp files on disk and the filenames returned. That seems like an extra step we shouldn’t have to do but I expect it might be a work-around for this problem.

mperrin:
Or, run two or more pools that each iterate only over part of the parameter space, so their returned data are smaller. Then combine together the outputs from the multiple pools. That’s probably simpler than adding in file output and input.

ivalaginja:
A Stackoverflow user suggested this was fixed in Python 3.8:
https://stackoverflow.com/a/53368158/10112569

I am looking into this now.

ivalaginja:
I set up an environment with Python 3.8 and get a different error message:

```bash
PAIR: 105-120
Calculating coro image...
Calculating mean contrast in dark hole
contrast: 2.3276463503564805e-10
Process SpawnPoolWorker-2:
Traceback (most recent call last):
  File "/Users/ilaginja/anaconda/envs/patest/lib/python3.8/multiprocessing/pool.py", line 131, in worker
    put((job, i, result))
  File "/Users/ilaginja/anaconda/envs/patest/lib/python3.8/multiprocessing/queues.py", line 368, in put
    self._writer.send_bytes(obj)
  File "/Users/ilaginja/anaconda/envs/patest/lib/python3.8/multiprocessing/connection.py", line 200, in send_bytes
    self._send_bytes(m[offset:offset + size])
  File "/Users/ilaginja/anaconda/envs/patest/lib/python3.8/multiprocessing/connection.py", line 395, in _send_bytes
    self._send(pre_header)
  File "/Users/ilaginja/anaconda/envs/patest/lib/python3.8/multiprocessing/connection.py", line 368, in _send
    n = write(self._handle, buf)
BrokenPipeError: [Errno 32] Broken pipe

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/ilaginja/anaconda/envs/patest/lib/python3.8/multiprocessing/process.py", line 313, in _bootstrap
    self.run()
  File "/Users/ilaginja/anaconda/envs/patest/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/ilaginja/anaconda/envs/patest/lib/python3.8/multiprocessing/pool.py", line 136, in worker
    put((job, i, (False, wrapped)))
  File "/Users/ilaginja/anaconda/envs/patest/lib/python3.8/multiprocessing/queues.py", line 368, in put
    self._writer.send_bytes(obj)
  File "/Users/ilaginja/anaconda/envs/patest/lib/python3.8/multiprocessing/connection.py", line 200, in send_bytes
    self._send_bytes(m[offset:offset + size])
  File "/Users/ilaginja/anaconda/envs/patest/lib/python3.8/multiprocessing/connection.py", line 404, in _send_bytes
    self._send(header)
  File "/Users/ilaginja/anaconda/envs/patest/lib/python3.8/multiprocessing/connection.py", line 368, in _send
    n = write(self._handle, buf)
BrokenPipeError: [Errno 32] Broken pipe
/Users/ilaginja/anaconda/envs/patest/lib/python3.8/multiprocessing/resource_tracker.py:203: UserWarning: resource_tracker: There appear to be 6 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```

I am dropping this, it's not worth investigating right now.

In case I pick this up at a later time, here are some notes:
- `jupyter` needs py < 3.8.0, which makes py3.8 essentially unusable for this repository since it relies heavily on jupyter notebooks. This would only be worth it if using py3.8 actually solves the above issue, which I don't know whether it does.
- to run PASTIS, I need `hcipy`, albeit still from an old commit that specifically requires `scipy` 1.2.1. This is not supported in py3.8, so there's another issue with that. I need to refactor the repo to use the most recent version of hcipy anyway (#4), which should fix that, but I am not there yet.
- if I still want to use hcipy from that specific commit, but in py3.8, I can keep the newer version of scipy, but need to manually add the following change in hcipy, plus remove the import of scipy at the top of that script:
https://github.com/ehpor/hcipy/commit/95dc7a661a4ddd0725def2a2ebf55e2024414c4d

I think the conclusion here is that I need to refactor the code for usage with hcipy and that I am dropping any multiprocessing for now.

ivalaginja:
Created a cleaner draft PR of this in #24, closing this without merging.

PR #24, opened 2019-11-27
Multiprocess PASTIS matrix

ivalaginja:
Cleaner copy of #22, see there for comments and notes.

Back at this after dusting it off:
- rebased on most recend dev (the version that's (almost) ready for a new release)
- adjusted the multiprocessed function to reflect the changes done in the main matrix function in the past 7 months:
  - implemented logging
  - saving out of configfile with data
  - updated output file names
  - renamed `wfe_aber`
  - convert all `print()`/`log.info()` parameters to either f strings or formatted strings
- save PSFs and WFE maps within the function across processes, so that they don't have to be held in memory
- add notebook that allows me to check matrices against each other
- call the intermediate result it the more obvious name `contrast_matrix`

I will hold off on merging into dev since I don't want to include this in the v2.0.0 release that goes with the paper, since I didn't have this when doing the analyses for the paper.

### DIFF
--- a/Jupyter Notebooks/LUVOIR/18_comparison_matrices_multiprocessed.ipynb
+++ b/Jupyter Notebooks/LUVOIR/18_comparison_matrices_multiprocessed.ipynb
@@ -1,0 +1,254 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Comparing normal matrix results vs. multiprocessed matrix results\n",
+    "\n",
+    "Making sure the multiprocessing spits out the correct results. Verified on the LUVOIR case."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Imports\n",
+    "import os\n",
+    "from astropy.io import fits\n",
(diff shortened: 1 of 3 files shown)


ivalaginja:
Summary from #22: I implemented multiprocessing successfully by example of `hicat.scripts.stroke_minimization.calculate_jacobian`; it worked find on 5 segments but errors out on the full data set, because the result the pool returns is too large. This seems to be fixed in Python 3.8 but the repo is not set up to work with that version and doing that would be a major headache.

Alternatively, @mperrin suggested to cut down the work done per multiprocessing pool:
> what if you replaced:
> `results = mypool.map(luvoir_matrix_pair, product(np.arange(nb_seg), np.arange(nb_seg)))`
>
> with instead:
> ```py
> results = []
> for i_seg in np.arange(nb_seg):
>   results.append(mypool.map(luvoir_matrix_pair, product(i, np.arange(nb_seg)))
> ```
> That way the amount done per multiprocessing pool is reduced by 1./nb_seg.

ivalaginja:
I now save the PSFs and WFE maps from all aberrated pairs directly in the function across processes, so that they don't have to be held in memory anymore. Only the mean contrast per aberrated pair, and tuple of the segment pair now get returned and there is no problem anymore.

PR #25, opened 2020-01-15
General cleanup

ivalaginja:
Changes:
- data directories now have a time stamp and instrument name, the "active" directory does not exist anymore (for LUVOIR; this will be in a hybrid state for other instruments now and will be fixed later)
- configfile now gets copied to output folder
- I modularized the code more
- I added a new script where I can mix and match what part of the analysis I run for LUVOIR
- added a new diagnostic notebook to double-check results
- added the comment about matrix normalization yet again, hopefully clear enough this time...
- general minor cleanups


ivalaginja:
Note how there is a keyword for the LUVOIR coronagraph model in the coffigfiel, which gets copied into the output directory, however this is NOT USED in any code directly. I am just using it for as a central switch for scripts where I owl have to change it in several places otherwise.

This is probably not the best way to do this, but I am pushing a decision and change for this to a later time.

PR #41, opened 2020-08-26
Implement HiCAT PASTIS analysis

ivalaginja:
Follow up to #40.

Rewrite the PASTIS analysis script such that it can do both HiCAT and LUVOIR.
Note: I am ignoring that the label positions in the plots will be wrong for the HiCAT results and hence won't show up. I will deal with this at a later point.

- [x] make the PASTIS analysis work for HiCAT
    - [x] create modes
    - [x] calculate mode constraints
    - [x] calculate cumulative contrast
    - [x] calculate segment constraints
    - [x] run Monte Carlo simulations
    - [x] plot results
    - [x] calculate covariance matrices
    - [x] calculate analytical analysis
    - [x] segment based error budget
- [x] implement instances of numpy's random number generator (`numpy.random.RandomState`) to avoid specified seed from HiCAT simulator
- [x] move hard-coded strings for LUVOIR optics from simulator script to configfile
- [x] move WFE RMS ranges in hockeystick for LUVOIR and HiCAT form hard-coded to configfile

### DIFF
--- a/pastis/config.ini
+++ b/pastis/config.ini
@@ -58,6 +58,9 @@ im_size_lamD_hcipy = 30
 [HiCAT]
 ; aberration for matrix calculation, in NANOMETERS
 calibration_aberration = 10.
+; log10 limits of PASTIS validity
+valid_range_lower = -1
+valid_range_upper = 3
 
 ; telescope
 nb_subapertures = 37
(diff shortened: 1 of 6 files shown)


ivalaginja:
Note how `numpy.random.RandomState` is considered legacy as of numpy v1.17.0:
https://numpy.org/doc/stable/reference/random/legacy.html

Instead, it has a thing called `numpy.random.Generator` now:
https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator

However, `RandomState` doesn't seem to be deprecated anytime soon I will leave it as is. If I ever replaces it with `Generator`, I will have to up the numpy requirement to > 1.16.0

PR #42, opened 2020-08-28
Copy DM maps and implement simple ymmetric matrix

ivalaginja:
In this small PR:
- [x] I make sure that the Boston DM maps used for the HiCAT coronagraph get saved together with the matrix output
- [ ] I rewrite the matrix calculation script to only calculate half of the matrix, including the diagonal, since it is symmetric. This should cut its calculation time by a factor of ~2


ivalaginja:
I opened this PR into the wrong branch...

PR #43, opened 2020-08-28
Implement simple symmetric matrix and copy HiCAT BostonDM maps

ivalaginja:
In this small PR:
- [x] I make sure that the Boston DM maps used for the HiCAT coronagraph get saved together with the matrix output
- [ ] I rewrite the matrix calculation script to only calculate half of the matrix, including the diagonal, since it is symmetric. This should cut its calculation time by a factor of ~2
- [x] save the contrast matrix as `pair-wise_contrasts.fits` rather than ravelled in `pair-wise_contrasts.txt`

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ directory structure is as follows:
 |      |-- OTE_images
 |          |-- opd[...].pdf                      # PDF images of each segment pair aberration in the pupil
 |          |-- ...
-|      |-- pair-wise_contrasts.txt:              # list of E2E DH average contrasts per aberrated segment pair
+|      |-- pair-wise_contrasts.fits:             # contrast matrix - E2E DH average contrasts per aberrated segment pair (only half of it since it is symmetric)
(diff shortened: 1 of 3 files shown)


ivalaginja:
Testing this on the HiCAT case, the calculation time has indeed decreased by a little less than a factor of two:

OLD:
Runtime for matrix_building_numerical.py/multiprocess: 758.0676171779633sec = 12.634460286299388min

NEW:
Runtime for matrix_building_numerical.py/multiprocess: 404.2263789176941sec = 6.737106315294901min

ivalaginja:
... but, the result seems to be different:

Before:
Screenshot 2020-08-28 at 15 11 35

After:
Screenshot 2020-08-28 at 15 11 27

I will have to investigate to see where the bug is. In the meantime I will split this PR in two so that I can merge the part that saves the HiCAT DM maps.

ivalaginja:
Continued in #45

PR #45, opened 2020-08-28
Implement simple symmetric matrix with non-repeating segment pairs

ivalaginja:
Continues where I found a discrepancy in #43.

Rewrite the matrix calculation script to only calculate half of the matrix, including the diagonal, since it is symmetric. This should cut its calculation time by a factor of ~2.

### DIFF
--- a/Jupyter Notebooks/HiCAT/19_comparison matrices less pairs.ipynb
+++ b/Jupyter Notebooks/HiCAT/19_comparison matrices less pairs.ipynb
@@ -37,7 +37,7 @@
    "source": [
     "# Define directories\n",
     "dir_standard = '2020-09-02T10-59-17_hicat'\n",
-    "dir_new = '2020-09-02T14-50-44_hicat'\n",
+    "dir_new = '2020-09-02T18-08-16_hicat'\n",
     "\n",
     "# Define matrix name\n",
     "matrix_name = 'PASTISmatrix_num_piston_Noll1.fits'\n",
--- a/Jupyter Notebooks/HiCAT/19_comparison matrices less pairs.ipynb
+++ b/Jupyter Notebooks/HiCAT/19_comparison matrices less pairs.ipynb
@@ -37,7 +37,7 @@
    "source": [
     "# Define directories\n",
     "dir_standard = '2020-09-02T10-59-17_hicat'\n",
-    "dir_new = '2020-09-02T11-31-56_hicat'\n",
+    "dir_new = '2020-09-02T14-50-44_hicat'\n",
     "\n",
     "# Define matrix name\n",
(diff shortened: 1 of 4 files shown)


ivalaginja:
It seems like the symmetrization of the matrix isn't quite accurate, I can see differences in the hockey stick curve when I use different methods for that:

Original hockey stick curve with full matrix being calculated:

Screenshot 2020-09-01 at 14 38 16


(1) When using `M = M + np.triu(M, 1)`:

Screenshot 2020-09-01 at 14 38 26


(2) When using `M = M + M.T - np.diag(M.diagonal())`:

Screenshot 2020-09-01 at 14 38 33

(2) reproduces the hockey stick curve less accurately than (1), although I would have expected those two methods to  yield exactly the same result. More conrcerning is though that none of them correctly reproduce the original result.

ivalaginja:
I need to remember that the coronagraph floor needs to be subtracted from the contrast matrix that comes fresh out of the simulator. The version of the contrast matrix that I save to disk does *not* have that done yet. This was done on purpose, so that I save the direct measurement to disk - I just have to remember to subtract the coro floor when I further use this matrix from disk.

ivalaginja:
I also need to remember that I can't just take the half-measurement matrix which contains the mean contrasts of all aberrated segment pairs and then just subtract the coro floor off that. This leads to the empty side of the matrix, which gets initialized with all zeros, being over-subtracted, since I don't symmetrize the matrix by replacing entries, but but adding them.

Solved with commit ad49077.

ivalaginja:
I might want to go back to this symmetrization function, since it doesn't depend on which part of the matrix is the one that needs to be "copied", the upper triangle or the lower triangle:

```py
def symmetrize(array):
    """
    Return a symmetrized version of NumPy array a.

    Values x are replaced by the array value at the symmetric
    position (with respect to the diagonal), i.e. if a_ij = x,
    then the returned array a' is such that a'_ij = a_ji.
    Diagonal values are left untouched.
    a -- square NumPy array, such that a_ij = x or a_ji = x,
    for i != j.

    Source:
    https://stackoverflow.com/a/2573982/10112569
    """
    return array + array.T - np.diag(array.diagonal())
```

ivalaginja:
I made sure the resulting PASTIS matrix is *exactly* the same in this PR as it is when I calculate it on dev. I used the notebook in this PR for this verification.

ivalaginja:
And the matrix calculation time goes down by about a factor of two indeed:

Original runtime:
```2020-09-02 11:44:42,828 - root - INFO - Runtime for matrix_building_numerical.py/multiprocess: 766.0100440979004sec = 12.76683406829834min```

New runtime:
```2020-09-02 18:14:55,871 - root - INFO - Runtime for matrix_building_numerical.py/multiprocess: 399.86494612693787sec = 6.664415768782297min```

PR #62, opened 2020-10-16
Apply DH mask to focal plane PASTIS mode plots

ivalaginja:


ivalaginja:
nope, this needs the things in #60 , so I will leave this one commit in there

PR #72, opened 2020-12-29
Write first tests for PASTIS matrix regression and basic PASTIS analysis

ivalaginja:
This PR adds some basic tests. The main one compares a newly calculated LUVOIR small PASTIS matrix against one previously calculated, that is tracked with Git LFS.

These tests are currently passing when run locally and when a `config_local.ini` is present, meaning the code can read the keys in the section `[local]`. I will make it work with the default configfile and set up CI with GitHub Actions in a follow-up PR.

I have also changed the main function for the calculation of the PASTIS matrix to take in the output path for results as a function parameter.


ivalaginja:
CI enabled for default configfile in #74

ivalaginja:
CI with GitHub Actions is being set up in #75

PR #75, opened 2020-12-29
Set up GitHub Actions

ivalaginja:
Setting up CI with GitHub Actions. Adapted the file CI.yml from @spacetelescope/hicat-package


ivalaginja:
All tests passing, but I will only run them on ubuntu in the future and not additionally on Windows since they take forever.

PR #80, opened 2021-01-09
Test and fix drifting coronagraph floor

ivalaginja:
Follow-up to #79 where I rewrote the analytical PASTIS matrix calculation to be able to also take in a variable contrast floor per measurement for the PASTIS matrix. There, I only made sure the changes are backwards compatible with cases in which we still assume a constant contrast floor.

In this PR I made small fixes to make sure the implementation for a drifting contrast floor are actually correct, and I add a test that checks the correctness of this calculation.


ivalaginja:
The tests in the analysis are probably failing due to the new PASTIS matrix used for the tests. The results are close but not as close as they used to be, so I will have to look into this more closely.

ivalaginja:
Ha, I pushed the matrices from a wrong simulation run, which were not correct. Good I have these tests I suppose.

ivalaginja:
I pushed two matrices (PASTIS matrix and contrast matrix) from a run generated on commit `2a9dd6`.
Tests are passing locally now.

ivalaginja:
There we go

PR #81, opened 2021-01-13
Add function for animation of contrast matrix generation

ivalaginja:
Previously I only had this in a notebook, I cleaned it up and put it in the `plotting` module.
Only LUVOIR data supported for now, but extending for HiCAT should be trivial.


ivalaginja:
Adding a progressbar would be great, as running this function takes forever...

PR #87, opened 2021-03-21
Extend segmented DM to arbitrary number of Zernikes, add global Zernike DM

ivalaginja:
Add functionality to the LUVOIR simulator to use the most recent hcipy and update its segmented deformable mirror to use hcipy.optics.DeformableMirror(), to be able to do an indefinite number of local Zernikes.
Also, add methods to create an active optic to create independent global Zernikes on the primary mirror, as well as a general deformable mirror.

Does NOT switch the simulator to using those yet in the scripted PASTIS matrix generation, or analysis.


ivalaginja:
Tests are failing because the new LUVOIR matrix is slightly different from the old one that's currently loaded in the tests. This is due to the upgrade to hcipy 0.4.0 and how the grids get created. This new version eliminates a centering error we previously had in the simulator, so the matrix in Git LFS should be replaced with the new one.

ivalaginja:
Tests pass locally, so I really just have to figure out why Git LFS won't track my new matrix files...

ivalaginja:
For the record, below the difference between the old and new PASTIS matrix for the small APLC on LUVOIR A. Colorbar units are contrast/nm^2.

Old coro_floor: 4.315823935036038e-11
New coro_floor: 4.2376360700565846e-11

Difference (old - new):
Screenshot 2021-03-21 at 22 12 46

Absolute difference:
Screenshot 2021-03-21 at 22 12 25

ivalaginja:
Thanks to the help of @jamienoss the matrix files are *actually* tracked by LFS now. Tests are passing, so I am merging.
Fyi @ananyasahoo1904 @lpueyo

PR #88, opened 2021-03-23
Add Harris and ripple DMs to simulator

ivalaginja:
Like in #87, this doesn't change the scripts that calculate the matrix or do the analysis, but this adds functionality to the simulator:
- create and use a DM putting Fourier modes on the primary mirror
- create and use a segmented DM that applies custom Harris modes to the primary mirror
I have implemented the Harris segmented mirror such that it can exist at the same time like the Zernike segmented mirror, in case we ever want to use both at the same time.

This should fully enable us to run @lpueyo's notebooks off of the default branch. Fyi @ananyasahoo1904


ivalaginja:
This works as expected, below you can see the output of the Jupyter notebook that was changed in this PR. Each deformable mirror is in the initial pupil plane and can be actuated independently. They are then combined for the propagation.

Screenshot 2021-03-23 at 22 09 03

PR #90, opened 2021-03-24
Split simulator classes

ivalaginja:
I am splitting up the simulator classes that we were using so far for LUVOIR (`SegmentedTelescopeAPLC` and `LuvoirAPLC`) into three new classes that inherit form each other sequencially:

- `SegmentedTelescope`
- `SegmentedAPLC(SegmentedTelescope)`
- `LuvoirA_APLC(SegmentedAPLC)`

Theh idea behind this is to make the `SegmentedTelescope` and `SegmentedAPLC` classes more general, looking ahead to implementations of other coronagraphs and specific observatories. Other coronagraphs can be implemented as e.g. `SegmentedVVC`, and a segmented telescope that uses a vortex coronagraph can then inherit from `SegmentedVVC`.

This PR keeps everything back-compatible; and it does not swap in the new classes for the old ones yet in spots where the simulator is used.

### DIFF
--- a/Jupyter Notebooks/LUVOIR/21_extend_LUVOIR_simulator.ipynb
+++ b/Jupyter Notebooks/LUVOIR/21_extend_LUVOIR_simulator.ipynb
@@ -176,7 +176,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# For usage with Zernike mode mirror\n",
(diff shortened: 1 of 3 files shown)


ivalaginja:
@ananyasahoo1904 @lpueyo the new simulator classes are merged now and should be used from now on instead of the old ones.

PR #93, opened 2021-03-26
Enable propagations normalized to one photon in the entrance pupil

ivalaginja:
Adds optional parameters to both `calc_psf()` methods, as well as `calc_out_of_band_wfs()` and `calc_low_order_wfs()` that normalize their propagations to one photon on the telescope aperture.

### DIFF
--- a/pastis/e2e_simulators/luvoir_imaging.py
+++ b/pastis/e2e_simulators/luvoir_imaging.py
@@ -434,13 +434,16 @@ class SegmentedTelescope:
     def _propagate_active_pupils(self, norm_one_photon=False):
         """ Propagate aperture wavefront "through" all active entrance pupil elements (DMs).
 
+        Parameters:
+        ----------
+        norm_one_photon : bool
+            Whether or not to normalize the returned E-fields and intensities to one photon in the entrance pupil.
+
         Returns:
         --------
         wf_active_pupil, wf_sm, wf_harris_sm, wf_zm, wf_ripples, wf_dm : hcipy.Wavefronts
             E-field after each respective DM individually; all DMs in the case of wf_active_pupil.
-        norm_one_photon : bool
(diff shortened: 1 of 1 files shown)


ivalaginja:
Since this was the last PR that dealt with the simulator upgrade, I'll put this summary here. The original function names are from Laurent's branch on which he developed his analyses.

**Luvoir simulator function mapping**

Some mirrors have changed names:
- self.dm stays the same
- self.zm —> self.zernike_mirror
- self.fm —> self.ripple_mirror
- self.sm stays the same when using local Zernikes, but —> self.harris_sm when for Harris modes
- self.zernike_wfs —> self.zwfs; and can use self.create_zernike_wfs() to create it, bu this happens automatically when using calc_low_order_wfs (unless it already exists)

Creating the mirrors:
- make_LO_Modes —> create_global_zernike_mirror
- make_DM —> create_continuous_deformable_mirror
- make_HI_Modes —> create_ripple_mirror
- make_segment_zernike_primary —> create_segmented_mirror
- make_segment_Harris_thermal_primary —> create_segmented_harris_mirror

Propagating WFS:
- prop_OBWFS —> calc_out_of_band_wfs
- prop_LOWFS —> calc_low_order_wfs

Any function that starts with “calc_” now also has the parameter “norm_one_photon” with can be True or False. The default is False, but when set to true, it will normalise the propagation to one photon. This is applicable for:
- calc_out_of_band_wfs
- calc_low_order_wfs
- calc_psf
There’s also the direct propagator that is normalised to one photon:
- prop_norm —> prop_norm_one_photon

Additionally, you can remove the various mirrors with the following methods. This helps speed up things when you’re not actually using a particular mirror:
- remove_segmented_mirror
- remove_segmented_harris_mirror
- remove_global_zernike_mirror
- remove_ripple_mirror
- remove_continuous_deformable_mirror

PR #96, opened 2021-05-09
Add LUVOIR B simulator

ivalaginja:
Add a LUVOIR B simulator with charge 6 Vortex coronagraph, initially. Starting by porting over the work @lpueyo did to prepare the telescope and coronagraph, then I am integrating it with the existing `SegmentedTelescope` functionalities.

This will likely see improvements in the future.

Before merging this in a usable state, I want to:
- [x] Revisit how we created the indexed aperture for LUVOIR A, to be able to use the lighter weight segmented mirror (PTT only).
- [x] Double check that the ripple mirror and continuous DM in the first pupil plane are working, although if they don't, I might push fixing this to later since they are not the most important ones for now.
- [x] Move the more specific child classes (Luvoir_APLC etc) out into a separate module, maybe. The main one is getting very large.

This PR makes the simple segmented mirror for piston, tip, tilt only, work. The arbitrary continuous DM, and the global Zernike DM work as well, as does the OBWFS.

---
Follow-up work needed:
- See whether I can **number the segments circularly**, to have a similar segment ordering like usually used in segmented pupils (starting in the center and then spiraling out). This might make some of the analysis and comparisons easier.
- Make the **ripple mirror** work. Although it doesn't depend on the segmented aperture, for some reason it's showing awkward behavior that I want to get to the ground of. It might me a typo somewhere.
- Make **general segmented DM** and **Harris local mode DM** work. These depend on exact segment center positions, which are currently only eye-balled, so they are not working correctly. This will likely take some fiddling with the 55 individual segments.


ivalaginja:
This now produces the correct coronagraphic image:

Screenshot 2021-05-17 at 16 31 18

PR #98, opened 2021-05-17
Use classes for matrix calculation

ivalaginja:
This PR packs the calculation of the numerical matrix into an object-oriented structure. I refactor the calculation of matrices from intensities, however I set it up in such a way that it should be fairly easy to write a new child class for matrices from E-fields. This is also written in such a way that it should become very easy to script up the matrix generation for other mirrors than the segmented mirror, e.g. low-order or Harris mode mirror.

This is a major change in terms of how the matrix generation is controlled in the main scripts and should facilitate the porting of other used cases to scripts that we can spin up on the servers, e.g. the custom segment modes (fyi @ananyasahoo1904).
FYI @lpueyo this is also what will allow us to just drop in any new telescope + coronagraph combo (e.g. LUVOIR B) to just use the same code in order to make matrices.

None of this touches any of the PASTIS analysis, which I will adjust separately.

Still to do:
- [ ] ~Split up the function `calculate_unaberrated_contrast_and_normalization()` in `matrix_building_numerical.py` into the according E2E simulator modules, and refactor usages accordingly; including the ones in the new matrix child classes. This should get rid of quite a number of very ugly `if...elif...else` blocks in which I differentiate between the simulators.~ --> This would actually break a lot of things, so let's not do that for now.
- [ ] ~Split up the module `matrix_building_numerical.py`. Ideas include creating a new subdirectory dedicated only to numerical matrix calculations, since there will be a significant proliferation between instruments and used mirrors. Move simulator-specific stuff into the respective simulator modules.~ --> Do this when I actually start implementing new subclasses.
- [x] Use `abc` to implement an abstract base class with abstract methods,
- [x] Rid the class `PastisMatrix` of the `instrument` input parameter.
- [x] test for small LUVOIR APLC, incl. analysis
- [x] test for medium LUVOIR APLC, incl. analysis
- [x] test for large LUVOIR APLC, incl. analysis
- [x] test for HiCAT, incl. analysis
- [x] test for JWST, incl. analysis


ivalaginja:
The HiCAT analysis is causing me headaches on this branch, so this will have to wait till tomorrow.

ivalaginja:
Ha, it just didn't flatten the IrisAO when it was calculating the contrast floor. Fixed now, seems to give correct results now.
I will let the offline HiCAT test finish, then push the updates and merge without running the CI again, since that passed and I haven't touched anything LUVOIR related in the meantime.

PR #100, opened 2021-05-27
Calculate simulated PASTIS matrices from E-fields

ivalaginja:
Implement a way to calculate simulated PASTIS matrices efficiently from E-fields rather than intensities. The recent refactor and modularization in #98 makes this much easier, and I am following a similar subclassing approach for this PR.

Here I am implementing for LUVOIR-A only, and only for the multi-mode segmented mirror that's based on hcipy. Making the used modes/DMs swappable will happen in a follow-up PR, as well as writing subclasses for other instruments.

### DIFF
--- a/pastis/matrix_building_numerical.py
+++ b/pastis/matrix_building_numerical.py
@@ -190,6 +190,8 @@ def _jwst_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, segment_pa
 
 def _luvoir_matrix_single_mode(number_all_modes, wfe_aber, luvoir_sim, resDir, saveefields, mode_no):
 
+    log.info(f'MODE NUMBER: {mode_no}')
+
     # Apply calibration aberration to used mode
     all_modes = np.zeros(number_all_modes)
     all_modes[mode_no] = wfe_aber / 2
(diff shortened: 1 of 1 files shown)


ivalaginja:
The PASTIS matrix for LUVOIR-A/small APLC with piston only on the multi-mode segmented DM turns out ok when calculated from E-fields instead of Intensities:
Screenshot 2021-05-28 at 12 31 28

The difference is on the order of 1e-12 overall:
Screenshot 2021-05-28 at 12 32 21

ivalaginja:
@lpueyo said that he observed the same sort of differences between the PASTIS matrix form intensity and from E-fields when he was initially implementing the E-field combinations. In the final analysis results, the differences between the two matrix calculation methods are that some high sensitivity modes swap order compared to the matrix from intensities, and the 2-3 most relaxed segment requirements gain some 3 pm in tolerance (120pm instead of 116/117pm). This is not a major change, but would justify to look at it in more detail at some point when somebody finds some time for it.

One candidate for making the difference might be the order of operations in the PASTIS matrix calculation:

In the intensity approach, we first normalize the measured contrasts by the calibration aberration, then we construct the half-matrix, then we symmetrize it. See:  
https://github.com/spacetelescope/PASTIS/blob/56f42f84026c0fa2e144153bf655a14e1201c761/pastis/matrix_building_numerical.py#L295

In the E-field approach, we first calculate the half-matrix from the E-fields, then we symmetrize it, and we normalize by the calibration aberration at the very end. See:  
https://github.com/spacetelescope/PASTIS/blob/355933052de31f4b1085fed4a7833d54db7e4164/pastis/matrix_building_numerical.py#L364

So the point at which we normalize the matrix might make a difference. In any case, this is close enough to move forward so I will merge this as soon as the tests pass.

PR #105, opened 2021-06-08
RST CGI implementation

davidbourgeois:
Add PASTIS matrix calculation for RST CGI instrument, from intensities. Make hockey stick simulation work.
PASTIS analysis not touched in this PR.

Caveat: works with the RST simulator from https://github.com/davidbourgeois/webbpsf/tree/David-CGI_DM-Branch, commit 468d36d4cd723cd0fd82021c0a44001068a2ed8f


davidbourgeois:
I fixe wrong rebase, main problems were inside matrix_building_numerical.py

pastis/pastis_analysis.py
```diff
@@ -155,6 +167,8 @@ def full_modes_from_themselves(instrument, pmodes, datadir, sim_instance, saving
             plt.subplot(8, 5, i + 1)
         if instrument == 'JWST':
             plt.subplot(6, 3, i + 1)
+        if instrument == 'RST':
+            plt.subplot(6, 3, i + 1)
```
ivalaginja: @davidbourgeois I know now what you mean with the error you get. These subplots are supposed to plot each pastis mode into a grid, which means that you need to have at least as many grid spots like modes. For RST, you have 48x48=2304 modes, but here you are only using 6x3=18 grid points.

Since the sheer amount of modes with the continuous mirrors is not really feasible to put into a single plot, we might want to skip this figure altogether when working on a continuous mirror. But we'll look at this later, when we start working on on the analysis.

pastis/pastis_analysis.py
```diff
@@ -132,6 +132,18 @@ def full_modes_from_themselves(instrument, pmodes, datadir, sim_instance, saving
             jwst_wavenumber = 2 * np.pi / (CONFIG_PASTIS.getfloat('JWST', 'lambda') / 1e9)   # /1e9 converts to meters
```
ivalaginja: It would be good if you could revert all changes made to this file, since we want to come back to the analysis and actually test it later on. I think we have talked about how you can revert them but still keep them for later use. We can talk about this offline if you want.

pastis/util.py
```diff
@@ -676,3 +676,10 @@ def find_repo_location(package='pastis'):
     :param package: string, name of package within the repository whose path we are looking for
     """
     return os.path.abspath(os.path.join(find_package_location(package), os.pardir))
+
+
+def continous_dm_coo(nbactuator, segment):
```
ivalaginja: Let's rename this function to something more obvious (what does "coo" mean?) and please add a docstring.

pastis/util.py
```diff
@@ -203,7 +203,7 @@ def get_segment_list(instrument):
     :param instrument: string, "HiCAT", "LUVOIR" or "JWST"
     :return: seglist, array of segment numbers (names! at least in LUVOIR and HiCAT case. For JWST, it's the segment indices.)
     """
-    if instrument not in ['LUVOIR', 'HiCAT', 'JWST']:
+    if instrument not in ['LUVOIR', 'HiCAT', 'JWST' , 'RST']:
```
ivalaginja: This is nit-picking, but please delete the space before the comma.

pastis/plotting.py
```diff
@@ -21,7 +21,7 @@
 from pastis.config import CONFIG_PASTIS
 from pastis.e2e_simulators.luvoir_imaging import LuvoirAPLC
 import pastis.e2e_simulators.webbpsf_imaging as webbpsf_imaging
-import pastis.util
+import pastis.util as util
```
ivalaginja: I would actually prefer if we keep this to simply `import pastis.util`. The reason is that many packages have a `util` module and importing them without an alias makes it easier to understand what is going on, and to avoid conflicts.

Also, since the only thing that's added in this script is the plotting of the mu map, which we don't have yet, lets' revert all changes in this script as well (see comment for `pastis_analysis.py`)

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -330,7 +361,6 @@ def _jwst_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, segment_pa
 
     return contrast, segment_pair
 
-
```
ivalaginja: Please reinstate this deleted blank line.

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -383,7 +413,6 @@ def _luvoir_matrix_one_pair(design, norm, wfe_aber, resDir, savepsfs, saveopds,
 
     return float(contrast), segment_pair
 
-
```
ivalaginja: Here too, please reinstate this deleted blank line.

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -31,6 +31,10 @@
 matplotlib.rc('image', origin='lower')
 matplotlib.rc('pdf', fonttype=42)
 
+def copy(self):
+    """ Make a copy of a wavefront object """
```
ivalaginja: Fix this docstring please.

pastis/e2e_simulators/webbpsf_imaging.py
```diff
@@ -34,7 +34,6 @@
 FLAT_DIAM = CONFIG_PASTIS.getfloat('JWST', 'flat_diameter') * u.m
 IM_SIZE_E2E = CONFIG_PASTIS.getint('numerical', 'im_size_px_webbpsf')
 
-
```
ivalaginja: There's a bunch of deleted blank lines in this script, please put them back.

pastis/config_pastis.ini
```diff
@@ -39,6 +39,37 @@ filter_name = F335M
 lambda = 3350
 ; wavelength in nanometers
 
+[RST]
+; In devellopement
```
ivalaginja: --> "development"

pastis/config_pastis.ini
```diff
@@ -39,6 +39,37 @@ filter_name = F335M
...
+; aberration for matrix calculation, in NANOMETERS
+calibration_aberration = 5.
+; log10 limits of PASTIS validity in nm WFE
+valid_range_lower = -2
+valid_range_upper = 2
+
+; telescope
+# nb_subapertures put square value too set good actuators number for DMs
```
ivalaginja: --> "# nb_subapertures is the total number of actuators on the continuous DM"

.gitignore
```diff
@@ -6,4 +6,5 @@ config_local.ini
 .DS_Store
 pastis.egg-info
 build/
-dist/
\ No newline at end of file
+dist/
+UNKNOWN.egg-info/
```
ivalaginja: Where did this come from? Is this from the `webbpsf` installation?

pastis/contrast_calculation_simple.py
```diff
@@ -391,6 +391,69 @@ def contrast_jwst_num(coro_floor, norm, matrix_dir, rms=50*u.nm):
...
 
 
+def contrast_rst_num(coro_floor, norm, matrix_dir, rms=50*u.nm):
+    """
+    Compute the contrast for a random segmented OTE misalignment on the RST simulator.
+
+    :param coro_floor: float, coronagraph contrast floor
+    :param norm: float, normalization factor for PSFs: peak of unaberrated direct PSF
+    :param matrix_dir: str, directory of saved matrix
+    :param rms: astropy quantity (e.g. m or nm), WFE rms (OPD) to be put randomly over the entire segmented mirror
```
ivalaginja: We don't have a segmented mirror on RST, please fix the docstring.

pastis/e2e_simulators/webbpsf_imaging.py
```diff
@@ -166,6 +162,32 @@ def set_up_nircam():
 
     return nircam, ote
 
+def set_up_cgi():
+    """
+    Return a configured instance of the CGI simulator on RST.
+
+    Sets up the Lyot stop and filter from the configfile, turns of science instrument (SI) internal WFE
```
ivalaginja: "turns of" --> "turns off"

pastis/e2e_simulators/webbpsf_imaging.py
```diff
@@ -166,6 +162,32 @@ def set_up_nircam():
...
+    :return: Tuple of CGI instance
+    """
+    webbpsf.setup_logging('ERROR')
+
+    #Set actuators numbesr
+    mode_in = CONFIG_PASTIS.get('RST', 'mode')
+    nbactuator = int(CONFIG_PASTIS.get('RST', 'nb_subapertures'))
+    nbactuator_in = int(np.sqrt(nbactuator))
+    if nbactuator_in**2 != nbactuator:
+        log.error('nb_subapertures in config_pastis.ini is not a square!')
```
ivalaginja: A square of what? Please add clarification to the error message. Also, I how about you actually raise an error if thsi actually happens, because there is not point in continuing to run anything. You can do:
```py
if nbactuator_in**2 != nbactuator:
    error_msg = f"The number of subapertures from config_pastis.ini is {nbactuator}, which is not the square of the actuators per row (={nbactuator_in})!"
    log.error(error_msg)
    raise ValueError(error_msg)
```

pastis/launchers/run_rst.py
```diff
@@ -0,0 +1,32 @@
+"""
```
ivalaginja, line 1: The CI is failing on some syntax error in this script that we will have to look at.
Edit: you have a double `from` in line 8, that's why it's failing.

pastis/launchers/run_rst.py
```diff
@@ -0,0 +1,32 @@
...
+
+if __name__ == '__main__':
+
+    # Generate the matrix
+    run_matrix = MatrixIntensityRST(initial_path=CONFIG_PASTIS.get('local', 'local_data_path'))
+    run_matrix.calc()
+    dir_run = run_matrix.overall_dir
+
+    # Alternatively, pick data location to run PASTIS analysis on
+    # dir_run = os.path.join(CONFIG_PASTIS.get('local', 'local_data_path'), '2020-08-26T00-00-00_jwst')
```
ivalaginja: "jwst" --> "rst"

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -434,6 +463,57 @@ def _hicat_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, segment_p
 
     return contrast, segment_pair
 
+def _rst_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, actuator_pair):
+    """
+    Function to calculate RST mean contrast of one aberrated segment pair in NIRCam; for num_matrix_multiprocess().
```
ivalaginja: Don't mention `num_matrix_multiprocess()`; also, no NIRCam on RST.

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -434,6 +463,57 @@ def _hicat_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, segment_p
...
+    if savepsfs:
+        filename_psf = f'psf_actuator_{actuator_pair[0]}-{actuator_pair[1]}'
+        hcipy.write_fits(psf, os.path.join(resDir, 'psfs', filename_psf + '.fits'))
+
+    # Plot segmented mirror WFE and save to disk
+    if saveopds:
+        opd_name = f'opd_actuator_{actuator_pair[0]}-{actuator_pair[1]}'
+        plt.clf()
+        plt.figure(figsize=(8, 8))
+        rst_cgi.dm1.display( what='opd', opd_vmax=wfe_aber, colorbar_orientation='horizontal', title='Aberrated segment pair')
```
ivalaginja: Delete the space before "what='opd'".

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -434,6 +463,57 @@ def _hicat_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, segment_p
...
+        plt.savefig(os.path.join(resDir, 'OTE_images', opd_name + '.pdf'))
+
+
```
ivalaginja: Delete one blank line here.

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -734,8 +817,25 @@ def calculate_ref_image(self, save_coro_floor=True, save_psfs=True):
...
+                                                                                                               save_coro_floor=save_coro_floor,
+                                                                                                               save_psfs=save_psfs,
+                                                                                                               outpath=outpath)
+
 
 if __name__ == '__main__':
 
         MatrixIntensityLuvoirA(design='small', initial_path=CONFIG_PASTIS.get('local', 'local_data_path')).calc()
-        #MatrixIntensityHicat(initial_path=CONFIG_PASTIS.get('local', 'local_data_path')).calc()
+        #MatrixIntensityHicat(initial_path=CONFIG_PASTIS.get('local', 'local_data_path')).calc()
```
ivalaginja: Add a blank line at the end of the script.

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -253,9 +263,30 @@ def calculate_unaberrated_contrast_and_normalization(instrument, design=None, re
         # Return the coronagraphic simulator (a tuple in the JWST case!)
         coro_simulator = jwst_sim
 
-    # Calculate coronagraph floor in dark hole
-    contrast_floor = util.dh_mean(coro_psf, dh_mask)
-    log.info(f'contrast floor: {contrast_floor}')
+        contrast_floor = util.dh_mean(coro_psf, dh_mask)
+        log.info(f'contrast floor: {contrast_floor}')
```
ivalaginja: The idea behind this block being outside of, and after all `if` statements for the instruments is that each if block should define a coronagraphic PSF and DH mask in the same format, so that we avoid duplicating code. The minimum I would ask you to do is to is to bring back this unindented block and delete duplications in the if blocks of LUVOIR, HiCAT and JWST.

Ideally, you would also write the if block for RST such that it can also calcualte the coro_floor in the same way, from a previously defined coro_psf and dh_mask.

ivalaginja (changes requested):
This is very clean already, and clearly it runs since we have seen your results.
I have requested quite some cosmetic changes to keep things uniform on the repo, and to keep to the intended scope of the PR. Let me know if you have any questions.

davidbourgeois:
I did change, you asked, but I must test when my VM will work

pastis/matrix_generation/matrix_building_numerical.py
davidbourgeois: For one time is not my mistake :p

davidbourgeois:
Tested :)

pastis/util.py
```diff
@@ -676,3 +676,15 @@ def find_repo_location(package='pastis'):
     :param package: string, name of package within the repository whose path we are looking for
     """
     return os.path.abspath(os.path.join(find_package_location(package), os.pardir))
+
+
+def seg_to_dm_xy(nbactuator, segment):
+    """
+    Change references, segment number to x,y DM coordinate
```
ivalaginja: Suggestion: "Convert single index of DM actuator to x|y Dm coordinates. This assumes the actuators to be arranged on a square grid."

pastis/util.py
```diff
@@ -676,3 +676,15 @@ def find_repo_location(package='pastis'):
     :param package: string, name of package within the repository whose path we are looking for
     """
     return os.path.abspath(os.path.join(find_package_location(package), os.pardir))
+
+
+def seg_to_dm_xy(nbactuator, segment):
```
ivalaginja: Suggestion: rename "nbactuator" to "nbactuator_total" or something similar, to be clearer.

pastis/util.py
```diff
@@ -676,3 +676,15 @@ def find_repo_location(package='pastis'):
...
+    """
+    Change references, segment number to x,y DM coordinate
+    nbactuator : number actuator in each line (both x or y in square one)
+    segment : segment number input
```
ivalaginja: --> "int, single-index actuator number on the DM, to be converted to x|y coordinate"

pastis/util.py
```diff
@@ -676,3 +676,15 @@ def find_repo_location(package='pastis'):
     :param package: string, name of package within the repository whose path we are looking for
     """
     return os.path.abspath(os.path.join(find_package_location(package), os.pardir))
+
+
+def seg_to_dm_xy(nbactuator, segment):
+    """
+    Change references, segment number to x,y DM coordinate
+    nbactuator : number actuator in each line (both x or y in square one)
```
ivalaginja: This is very unclear. How about: "int, total number of actuators on (square) DM"
Also, English doesn't use spaces before colons (= ":") ^^

pastis/contrast_calculation_simple.py
```diff
@@ -391,6 +391,69 @@ def contrast_jwst_num(coro_floor, norm, matrix_dir, rms=50*u.nm):
     return contrast_jwst, contrast_matrix
 
 
+def contrast_rst_num(coro_floor, norm, matrix_dir, rms=50*u.nm):
+    """
+    Compute the contrast for a random segmented OTE misalignment on the RST simulator.
```
ivalaginja: Still don't have a segmented mirror on RST ^^ How about: "Compute the contrast for a random aberration over all DM actuators in the RST simulator."

pastis/e2e_simulators/webbpsf_imaging.py
ivalaginja: "turns of" still needs to be fixed.

pastis/launchers/run_rst.py
ivalaginja: This one ("jwst") still needs fixing.

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -435,6 +464,57 @@ def _hicat_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, segment_p
     return contrast, segment_pair
 
 
+def _rst_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, actuator_pair):
+    """
+    Function to calculate RST mean contrast of one aberrated segment pair in CGI.
```
ivalaginja: Let's call it "DM actuator pair" instead of "segment"

.gitignore
ivalaginja: Do you know why this file appears on the repo?

pastis/analytical_pastis/calibration.py
```diff
@@ -41,7 +41,7 @@
     outer_wa = CONFIG_PASTIS.getint(telescope, 'OWA')
     sampling = CONFIG_PASTIS.getfloat(telescope, 'sampling')
 
-    if telescope == 'JWST':
+    if telescope == 'JWST' or 'RST':
```
ivalaginja: I think you can actually revert this since we are not calculating the matrix analytically, so no need to touch this script at all.

pastis/config_pastis.ini
```diff
@@ -8,8 +8,8 @@ webbpsf_data_path = /Users/<user-name>/anaconda/envs/astroconda/share/webbpsf-da
 local_data_path = /Users/<user-name>/data_from_repos/pastis_data
 
 [telescope]
-name = LUVOIR
-; JWST, ATLAST, HiCAT or LUVOIR
+name = RST
+; JWST, RST, ATLAST, HiCAT or LUVOIR
```
ivalaginja: Keep the addition of "RST" in the comment, but let's leave the default for `name=` to be LUVOIR. I don't think this is actually being used anywhere, and if it does, it should be set in the local configfile, not in the main one. It's on my plate to actually revisit the usage of some of these configfile parameters.

pastis/contrast_calculation_simple.py
```diff
@@ -218,7 +218,7 @@ def contrast_hicat_num(coro_floor, norm, matrix_dir, rms=1*u.nm):
     log.info('Calculating E2E contrast...')
     # Put aberration on Iris AO
     for nseg in range(nb_seg):
-        hicat_sim.iris_dm.set_actuator(nseg, aber[nseg], 0, 0)
+        hicat_sim.iris_dm.set_actuator(nseg, aber[nseg]*u.nm, 0, 0)
```
ivalaginja: Did you test this for HiCAT? If not (takes too much time), you don't have to, but then it would be better if instead of adding `*u.nm` blindly to add a comment in that same line: "# TODO: likely misses a '*u.nm' for the correct conversion of the aberration."

pastis/contrast_calculation_simple.py
```diff
@@ -391,6 +391,69 @@ def contrast_jwst_num(coro_floor, norm, matrix_dir, rms=50*u.nm):
...
+    log.info(f'E2E: {end_e2e-start_e2e}sec = {(end_e2e-start_e2e)/60}min')
+    log.info(f'Matrix PASTIS: {end_matrixpastis-start_matrixpastis}sec = {(end_matrixpastis-start_matrixpastis)/60}min')
+
+    end_time = time.time()
+    runtime = end_time - start_time
+    log.info(f'Runtime for contrast_calculation_simple.py: {runtime} sec = {runtime/60} min')
+
+    return contrast_rst, contrast_matrix
+
 if __name__ == '__main__':
```
ivalaginja, line 464: Keep two spaces before the main block please.

pastis/contrast_calculation_simple.py
```diff
@@ -391,6 +391,69 @@ def contrast_jwst_num(coro_floor, norm, matrix_dir, rms=50*u.nm):
...
+
+    rst_sim = webbpsf_imaging.set_up_cgi()
+    rst_sim.fpm = CONFIG_PASTIS.get('RST', 'fpm')
+    nb_actu = rst_sim.nbactuator
+
+    log.info('Calculating E2E contrast...')
+    # Put aberration on OTE
+    rst_sim.dm1.flatten()
+    for nseg in range(total_seg):
+        actu_x , actu_y = util.seg_to_dm_xy(nb_actu, nseg)
```
ivalaginja: Delete the space after "actu_x", before the comma

pastis/contrast_calculation_simple.py
```diff
@@ -391,6 +391,69 @@ def contrast_jwst_num(coro_floor, norm, matrix_dir, rms=50*u.nm):
     return contrast_jwst, contrast_matrix
 
 
+def contrast_rst_num(coro_floor, norm, matrix_dir, rms=50*u.nm):
```
ivalaginja, line 394: The parameter `norm` is in fact not used in this function, which prompted me to check the CGI function `raw_psf()` in your fork of `webbpsf`:
https://github.com/davidbourgeois/webbpsf/blob/468d36d4cd723cd0fd82021c0a44001068a2ed8f/webbpsf/roman.py#L895

I can see that you normalized your coronagraphic image in there. While this is correct considering the result it returns, I think we should simplify this. For one because good code is modular code, which means we want to split this into smaller units that end up working together. For two because it will keep all the `contrast_[instrument]_num()` functions consistent with each other, which will make future work easier. And third because it will make the code run faster by avoiding to recalculate the direct PSF each time.

This will need a little bit of work in your CGI simulator, but the changes should be fairly easy to make:
If you want to use convenience functions, make one that only returns the normalization factor from the direct PSF (`norm`). If you *also* want to keep the function `raw_contrast()` that's fine, but we should not use it here because it breaks the pattern of the functions for the other instruments.

I believe the changes you need to do in this function here boil down to
- calculate the direct PSF and `norm` factor *outside* of this function, before you call it in the parent script, and pass `norm` into the function here
- calculate only the coronagraphic PSF inside this function, which means you will need to replace the usage of `rst_sim.raw_contrast()`
- you already have a class attribute in CGI that is the dark hole mask, use that with `util.dh_mean()` to calculate the raw contrast

The big advantage of this is that we do not keep recalculating the direct PSF each time you want a coronagraphic PSF - this should speed up your scripts by a factor of 2!

pastis/util.py
```diff
@@ -203,7 +203,7 @@ def get_segment_list(instrument):
     :param instrument: string, "HiCAT", "LUVOIR" or "JWST"
     :return: seglist, array of segment numbers (names! at least in LUVOIR and HiCAT case. For JWST, it's the segment indices.)
     """
-    if instrument not in ['LUVOIR', 'HiCAT', 'JWST']:
+    if instrument not in ['LUVOIR', 'HiCAT', 'JWST', 'RST']:
         raise ValueError('The instrument you requested is not implemented. Try with "LUVOIR", "HiCAT" or "JWST" instead.')
```
ivalaginja: Add "RST" to the list of supported instruments in the error message.

pastis/e2e_simulators/webbpsf_imaging.py
```diff
@@ -166,6 +163,34 @@ def set_up_nircam():
 
     return nircam, ote
 
+def set_up_cgi():
```
ivalaginja, line 170: Keep two blank lines before each new function definition - here, as well as further above in the script where they have been deleted.

pastis/e2e_simulators/webbpsf_imaging.py
ivalaginja: Also, please add a sentence that says that the FPM setting will also be read from the configfile (no FPM, or, a specific FPM). E.g. "FPM setting (out of beam = 'OFF', or a specific FPM name) is also read from the configfile."

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -734,6 +817,23 @@ def calculate_ref_image(self, save_coro_floor=True, save_psfs=True):
                                                                                                                save_psfs=save_psfs,
                                                                                                                outpath=self.overall_dir)
 
+class MatrixIntensityRST(PastisMatrixIntensities):
```
ivalaginja, line 818: Two blank lines before a new class please.

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -734,6 +817,23 @@ def calculate_ref_image(self, save_coro_floor=True, save_psfs=True):
...
+    def setup_one_pair_function(self):
+        self.calculate_matrix_pair = functools.partial(_rst_matrix_one_pair, self.norm, self.wfe_aber, self.resDir,
+                                                       self.savepsfs, self.saveopds)
+
+    def calculate_ref_image(self, save_coro_floor=False, save_psfs=False, outpath=''):
+        self.contrast_floor, self.norm, self.coro_simulator = calculate_unaberrated_contrast_and_normalization('RST',
+                                                                                                               return_coro_simulator=True,
+                                                                                                               save_coro_floor=save_coro_floor,
+                                                                                                               save_psfs=save_psfs,
+                                                                                                               outpath=outpath)
```
ivalaginja: This would be better with `outpath=self.overall_dir`, like in `MatrixIntensityLuvoirA()`, and then you can delete the `outpath` input parameter. This is only for the one single ref image (well, actually a pair of direct + coro), so it's fine (and actually advantageous) to save it directly in the overall output folder of a simulation run.

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -435,6 +464,57 @@ def _hicat_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, segment_p
...
+    # Set up RST simulator in coronagraphic state
+    rst_cgi = webbpsf_imaging.set_up_cgi()
+
+    # Put aberration on correct segments. If i=j, apply only once!
+    log.info(f'PAIR: {actuator_pair[0]}-{actuator_pair[1]}')
+
+    # Transform segment to actuators coordonates
+    nb_actu = rst_cgi.nbactuator
+    actu_i_x , actu_i_y = util.seg_to_dm_xy(nb_actu, actuator_pair[0])
+    actu_j_x , actu_j_y = util.seg_to_dm_xy(nb_actu, actuator_pair[1])
```
ivalaginja: Delete the spaces before the commas.

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -435,6 +464,57 @@ def _hicat_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, segment_p
...
+    :return: contrast as float, and segment pair as tuple
+    """
+
+    # Set up RST simulator in coronagraphic state
+    rst_cgi = webbpsf_imaging.set_up_cgi()
+
+    # Put aberration on correct segments. If i=j, apply only once!
+    log.info(f'PAIR: {actuator_pair[0]}-{actuator_pair[1]}')
+
+    # Transform segment to actuators coordonates
```
ivalaginja: --> "Transform single-actuator index to x|y coordinate on DM"

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -435,6 +464,57 @@ def _hicat_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, segment_p
...
+
+
+    # Put aberration on correct segments. If i=j, apply only once!
+    rst_cgi.dm1.flatten()
+    rst_cgi.dm1.set_actuator(actu_i_x, actu_i_y, wfe_aber)
+    if actuator_pair[0] != actuator_pair[1]:
+        rst_cgi.dm1.set_actuator(actu_j_x, actu_j_y, wfe_aber)
+
+    log.info('Calculating coro image...')
+    image = rst_cgi.calc_psf(nlambda=1, fov_arcsec=1.6)
```
ivalaginja: Hard-coded numbers need to *at least* be commented! Where does the number for `fov_arcsec` come from? Either put it in the configfile, or at least add a comment here.

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -435,6 +464,57 @@ def _hicat_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, segment_p
...
+    if savepsfs:
+        filename_psf = f'psf_actuator_{actuator_pair[0]}-{actuator_pair[1]}'
+        hcipy.write_fits(psf, os.path.join(resDir, 'psfs', filename_psf + '.fits'))
+
+    # Plot segmented mirror WFE and save to disk
+    if saveopds:
+        opd_name = f'opd_actuator_{actuator_pair[0]}-{actuator_pair[1]}'
+        plt.clf()
+        plt.figure(figsize=(8, 8))
+        rst_cgi.dm1.display(what='opd', opd_vmax=wfe_aber, colorbar_orientation='horizontal', title='Aberrated segment pair')
```
ivalaginja: "segment" --> "actuator"

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -435,6 +464,57 @@ def _hicat_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, segment_p
...
+        plt.savefig(os.path.join(resDir, 'OTE_images', opd_name + '.pdf'))
+
+    contrast = rst_cgi.raw_contrast()
```
ivalaginja: Using `raw_contrast()` here is ok, however I would discourage the usage of this function inside of pastis; see my main review comment.

pastis/matrix_generation/matrix_building_numerical.py
ivalaginja: Ok, I have made up my mind about this now, see my main review comment.
In a nutshell for here, please do not use `raw_contrast()`, but instead use more modular functions that allow you to get the `norm` factor, the coronagraphic PSF, and the DH mask individually. Then you can also reinstate the original order of this block of code to calculate the contrast in this function, with the same block being used for all instruments.

ivalaginja (changes requested):
Ok - lots of minor comments and formatting again, but this time I have one major change request: please do not use the function `raw_contrast()`.

The reason is that this function implicitly does a bunch of things that we need to be able to do separately. Those are:
1. calculate an unaberrated, **direct PSF**, and get the peak as the normalization factor `norm`
2. calculate a coronagraphic PSF
3. provide a DH mask
4. use the three things above to calculate the mean DH contrast

The reason why I want these things to be separate is because I don't want to recalculate the direct PSF every time I calculate a coronagraphic PSF. This is a huge waste of time since the direct PSF never changes, which means it is much easier to pass the normalization factor around, rather then to recalculate it each time.

This change should be fairly easy to do, since you already have all the functions and methods to calculate a direct or coronagraphic PSF, and you also already have an attribute that saves the DH mask in the CGI instance. This means that all you have to do is to replace all usages of `raw_contrast()` in the scripts on the pastis repository. And then you can use the internal function `util.dh_mean()` to calculate the mean contrast without recalculating the direct PSF each time (that's why this function exists).

Let me know if you have any questions. Otherwise, it's nice to see how clean and structured this PR is, well done!

davidbourgeois:
I must try change in  .gitignore 
fov_arcsec is parameter to zoom. the 1.6 factor was suggest inside webbpsf for CGI, with mask a zoom do better display (remove black band)
But all change are OK.

ivalaginja (approved):
Good to go, next steps are the E-field implementation and the analysis itself.

PR #106, opened 2021-06-15
Fix dh_mask in RST code

davidbourgeois:
Replace the default circular DH mask calculated from `pastis.util.create_dark_hole()` with the actual DH mask contained in the CGI simulator, currently a double-sided wedge DH.

### DIFF
--- a/pastis/contrast_calculation_simple.py
+++ b/pastis/contrast_calculation_simple.py
@@ -434,7 +434,8 @@ def contrast_rst_num(coro_floor, norm, matrix_dir, rms=50*u.nm):
     psf = image[0].data / norm
 
     # Get the mean contrast
-    dh_mask = util.create_dark_hole(psf, iwa, owa, sampling)
+    rst_sim.working_area(im=psf, inner_rad=iwa, outer_rad=owa)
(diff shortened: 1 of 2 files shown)


davidbourgeois:
Fixed and tested.

ivalaginja (approved):
It's good you found this, thanks for fixing it.
While at it, I looked at the function `working_area()`:
https://github.com/davidbourgeois/webbpsf/blob/468d36d4cd723cd0fd82021c0a44001068a2ed8f/webbpsf/roman.py#L877

It is rather inefficient to compute a PSF every time you want to calculate the DH mask, especially if you do that only to pull some data from the header. I would suggest you eliminate that. One way of doing that is to make the variables `pix_scale` and `lambdaD_scale` input parameters to the function.

In any case, that's a change on the webbpsf side, so it's not holding up this PR here, which can be merged as soon as the CI tests pass. I definitely urge you to make the changes mentioned above in some way, it will save you a lot of overheads.

davidbourgeois:
Yes, I expected PASTIS take dh_mask in input. If I can I would like rework to get just one time the dh_mask.

PR #108, opened 2021-06-21
Implement E-field matrix generation for RST

davidbourgeois:
~Improve dh_mask for RST intensity~  --> will be moved to different PR
Implement matrix generation from E-fields for RST
Add some documentation


pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -289,7 +289,7 @@ def calculate_unaberrated_contrast_and_normalization(instrument, design=None, re
         ppl.plot_direct_coro_dh(direct_psf, coro_psf, dh_mask, outpath)
 
     if return_coro_simulator:
-        return contrast_floor, norm, coro_simulator
+        return contrast_floor, norm, dh_mask, coro_simulator
```
ivalaginja: This change is not back-compatible so we will have to see how to deal with it. Let's talk offline about this.

pastis/config_pastis.ini
```diff
@@ -40,16 +40,14 @@ lambda = 3350
...
 ; aberration for matrix calculation, in NANOMETERS
 calibration_aberration = 5.
 ; log10 limits of PASTIS validity in nm WFE
 valid_range_lower = -2
 valid_range_upper = 2
 
 ; telescope
-# nb_subapertures is the total number of actuators on the continuous DM
-# nb_subapertures must be a square value
+# nb_subapertures put square value too set good actuators number for DMs
```
ivalaginja: I think the original first line commend you had here was good, keep it :)

pastis/matrix_generation/matrix_building_numerical.py
```diff
@@ -458,15 +458,15 @@ def _hicat_matrix_one_pair(norm, wfe_aber, resDir, savepsfs, saveopds, segment_p
...
+def _rst_matrix_one_pair(norm, dh_mask, wfe_aber, resDir, savepsfs, saveopds, actuator_pair):
     """
     Function to calculate RST mean contrast of one DM actuator pair in CGI.
     :param norm: float, direct PSF normalization factor (peak pixel of direct PSF)
     :param wfe_aber: calibration aberration per segment in m
     :param resDir: str, directory for matrix calculations
     :param savepsfs: bool, if True, all PSFs will be saved to disk individually, as fits files
     :param saveopds: bool, if True, all pupil surface maps of aberrated segment pairs will be saved to disk as PDF
-    :param actuator_pair:
+    :param actuator_pair: pair of actuator where is injected aberration
```
ivalaginja: It is always good to specify the type in the docstring, e.g.: `tuple, pair of actuators to aberrate. If same segment gets passed in both tuple entries, the actuator will be aberrated only once.`

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -20,11 +21,12 @@
 
 
 class PastisMatrixEfields(PastisMatrix):
-    instrument = None
 
     def __init__(self, design=None, initial_path='', saveefields=True, saveopds=True):
         super().__init__(design=design, initial_path=initial_path)
 
+        self.instrument = CONFIG_PASTIS.get('telescope', 'name')
+
```
ivalaginja: This is a change we cannot do because it will break things, so please revert. The `instrument` attribute gets set in the child class, exactly like you have it in `MatrixEfieldRST`, so there is no need to change anything.

Just revert this and everything should work as expected.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -88,14 +90,19 @@ def pastis_matrix_from_efields(electric_fields, efield_ref, direct_norm, dh_mask
...
     matrix_pastis_half = np.zeros([nb_modes, nb_modes])
 
     for pair in util.segment_pairs_non_repeating(nb_modes):
-        intensity_im = np.real((efields[pair[0]].electric_field - efield_ref) * np.conj(efields[pair[1]].electric_field - efield_ref))
+        if instrument == 'RST':
+            intensity_im = np.real(
+                (efields[pair[0]].wavefront - efield_ref) * np.conj(efields[pair[1]].wavefront - efield_ref))
+        elif instrument == 'LUVOIR':
+            intensity_im = np.real(
+                (efields[pair[0]].electric_field - efield_ref) * np.conj(efields[pair[1]].electric_field - efield_ref))
```
ivalaginja: We should try to find a more elegant solution for this. My suggestion would be to make the input parameter `efields` already carry the actual E-fields, rather than some third-party object instance. For this too, let's talk offline.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -134,6 +141,44 @@ def setup_single_mode_function(self):
         self.calculate_one_mode = functools.partial(_luvoir_matrix_single_mode, self.number_all_modes, self.wfe_aber,
                                                     self.luvoir, self.resDir, self.save_efields, self.saveopds)
 
+class MatrixEfieldRST(PastisMatrixEfields):
```
ivalaginja, line 139: One more blank line before the class definition please ^^

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -134,6 +141,44 @@ def setup_single_mode_function(self):
...
+    """
+    Class to calculate RST Electrical field of RST CGI.
+    """
+    instrument = 'RST'
+
+    def __init__(self, max_local_zernike=3, initial_path='', saveefields=True, saveopds=True):
+        super().__init__(initial_path=initial_path, saveefields=saveefields, saveopds=saveopds)
+        self.max_local_zernike = max_local_zernike
```
ivalaginja: You don't need this input parameter or attribute (`max_local_zernike`) in the RST case, so delete all three of them.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -134,6 +141,44 @@ def setup_single_mode_function(self):
...
+        self.norm = direct_psf.max()
+
+        # Calculate dark hole mask
+        self.rst_cgi.working_area(im=direct_psf, inner_rad=iwa, outer_rad=owa)
+        self.dh_mask = self.rst_cgi.WA
+
+        # Calculate reference E-field in focal plane, without any aberrations applied
+        _trash, inter = self.rst_cgi.calc_psf(nlambda=1, fov_arcsec=1.6, return_intermediates=True)
+        self.efield_ref = inter[6].wavefront
+        # [6] is the last optic = detector
```
ivalaginja: Please move this comment to the line it belongs to, since it seems to be an in-line comment.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -134,6 +141,44 @@ def setup_single_mode_function(self):
...
+
+    def setup_deformable_mirror(self):
+        """DM setup not needed for RST, just define number of total mode"""
```
ivalaginja: "total number of mode**s**" (the "s" is missing)

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -160,3 +205,47 @@ def _luvoir_matrix_single_mode(number_all_modes, wfe_aber, luvoir_sim, resDir, s
...
+def _rst_matrix_single_mode(wfe_aber, rst_sim, resDir, saveefields, saveopds, mode_no):
+    '''
+    Function to calculate RST Electrical field (E_field) of one DM actuator in CGI.
+    :param wfe_aber: float, amplitude injected inside the chosen actuator
+    :param rst_sim: full optic CGI instance
+    :param resDir: str, directory for matrix calculations
+    :param saveefields: bool, if True, all E_field will be saved to disk individually, as fits files
+    :param savepods: bool, if True, all pupil surface maps of aberrated segment pairs will be saved to disk as PDF
+    :param mode_no: int, actif actuator where is injected aberration
+
```
ivalaginja: delete blank line

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -160,3 +205,47 @@ def _luvoir_matrix_single_mode(number_all_modes, wfe_aber, luvoir_sim, resDir, s
         plt.savefig(os.path.join(resDir, 'OTE_images', opd_name + '.pdf'))
 
     return efield_focal_plane
+
+
+def _rst_matrix_single_mode(wfe_aber, rst_sim, resDir, saveefields, saveopds, mode_no):
+    '''
```
ivalaginja: Nit-picking, but please use " instead of ' for docstrings.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -160,3 +205,47 @@ def _luvoir_matrix_single_mode(number_all_modes, wfe_aber, luvoir_sim, resDir, s
...
+    Function to calculate RST Electrical field (E_field) of one DM actuator in CGI.
+    :param wfe_aber: float, amplitude injected inside the chosen actuator
+    :param rst_sim: full optic CGI instance
```
ivalaginja: --> "instance of CGI simulator"

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -160,3 +205,47 @@ def _luvoir_matrix_single_mode(number_all_modes, wfe_aber, luvoir_sim, resDir, s
         plt.savefig(os.path.join(resDir, 'OTE_images', opd_name + '.pdf'))
 
     return efield_focal_plane
+
+
+def _rst_matrix_single_mode(wfe_aber, rst_sim, resDir, saveefields, saveopds, mode_no):
+    '''
+    Function to calculate RST Electrical field (E_field) of one DM actuator in CGI.
+    :param wfe_aber: float, amplitude injected inside the chosen actuator
```
ivalaginja: **Always** specify input units --> "float, calibration aberration per actuator in m" (correct the m if that is not true)

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -160,3 +205,47 @@ def _luvoir_matrix_single_mode(number_all_modes, wfe_aber, luvoir_sim, resDir, s
...
+    :param rst_sim: full optic CGI instance
+    :param resDir: str, directory for matrix calculations
+    :param saveefields: bool, if True, all E_field will be saved to disk individually, as fits files
+    :param savepods: bool, if True, all pupil surface maps of aberrated segment pairs will be saved to disk as PDF
```
ivalaginja: -> "aberrated actuators"

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -160,3 +205,47 @@ def _luvoir_matrix_single_mode(number_all_modes, wfe_aber, luvoir_sim, resDir, s
...
+    :param mode_no: int, actif actuator where is injected aberration
```
ivalaginja: --> "int, which aberrated actuator to calculate the E-field for"

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -160,3 +205,47 @@ def _luvoir_matrix_single_mode(number_all_modes, wfe_aber, luvoir_sim, resDir, s
...
+
+    '''
+
+    log.info(f'MODE NUMBER: {mode_no}')
```
ivalaginja: maybe call this `f'ACTUATOR NUMBER:...'` instead

ivalaginja (changes requested):
This looks good except for a a couple small things I asked for changes for.
Could you address all the comments *except* the ones about `calculate_semi_analytic_pastis_from_efields` and the passed in DH mask for the intensity case, and then let me know? Then we can talk about how to address those two things.

pastis/matrix_generation/matrix_building_numerical.py
davidbourgeois: May be we can do the same change for all coronagraphs? It will be better for performances.

ivalaginja: It probably would, but testing all cases needs a lot of time. I think the easiest, although not the best, path forward is to revert this to what it was, and make a note to change it at a later point. In any case, such intrusive refactors are better done as the only action point in a separate PR, to make sure nothing goes wrong.

ivalaginja: @davidbourgeois I encourage you to make a new branch in which you keep these changes, and only these changes. Then we can use that branch to do the full refactor and test it.

pastis/matrix_generation/matrix_from_efields.py
ivalaginja: @davidbourgeois you still need the `def __init__()` ! 😱  I mean to delete all three instances of `max_local_zernike`, not all three lines.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -135,6 +142,40 @@ def setup_single_mode_function(self):
                                                     self.luvoir, self.resDir, self.save_efields, self.saveopds)
 
 
+class MatrixEfieldRST(PastisMatrixEfields):
+    """
+    Class to calculate RST Electrical field of RST CGI.
```
ivalaginja: Just found this - this class calculates the PASTIS matrix from E-fields, not just the E-fields themselves.

pastis/matrix_generation/matrix_from_efields.py
ivalaginja: @davidbourgeois this still needs to be addressed.

pastis/matrix_generation/matrix_from_efields.py
davidbourgeois: Yes, I understand but, worked. I correct it

ivalaginja: Yeah I just realized that even without it would just use the inherited `__init__()` method, which is correct, but I think it's worth to keep things consistent between the individual child classes. Refactors should happen separately from new features, to avoid bugs. Thanks for fixing it!

davidbourgeois: I can't commit it...

ivalaginja: Why not, what's happening?

davidbourgeois: A stash blocked tree. I removed it, works now.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -95,7 +96,7 @@ def calculate_semi_analytic_pastis_from_efields(efields, efield_ref, direct_norm
     matrix_pastis_half = np.zeros([nb_modes, nb_modes])
 
     for pair in util.segment_pairs_non_repeating(nb_modes):
-        intensity_im = np.real((efields[pair[0]].electric_field - efield_ref) * np.conj(efields[pair[1]].electric_field - efield_ref))
+        intensity_im = np.real((efields[pair[0]] - efield_ref) * np.conj(efields[pair[1]] - efield_ref))
```
ivalaginja, line 99: @davidbourgeois the change you made in the return of `_luvoir_matrix_single_mode()` is exactly what was needed here to make it more streamlined 👍 This means that the entries of `efields` and `efield_ref` simply have to have the same type, but they're not tied to a specific type.

Now, does this work ok for RST, did you test this?

ivalaginja: @davidbourgeois confirmed that this works indeed

ivalaginja (approved):

pastis/matrix_generation/matrix_from_efields.py
davidbourgeois: I am @davidbourgeois and I confirm it.
----> OK I go out...

PR #110, opened 2021-06-25
Optimize dh_mask performances

davidbourgeois:
Compute one time per telescope dh_mask, and remove it from _telescope_one_pair()


davidbourgeois:
LUVOIR size matrix change too often.

PR #112, opened 2021-08-03
Telescope class and generalization

davidbourgeois:
Add parameters inside config
Automatize launcher
Generalize functions
Create Telescope class


davidbourgeois:
Sorry is a bit loud... I will do a different PR to generalize analysis.

PR #117, opened 2021-09-20
Add calibration amplitude as input parameter to pastis matrix

asahooexo:
In this PR,  add an extra input parameter to for pastis matrix generation. Currently, by default the aberration is 1nm. Other user should able to change this as per their requirements (say to 1pm or 10nm etc).

### DIFF
--- a/pastis/launchers/run_luvoir.py
+++ b/pastis/launchers/run_luvoir.py
@@ -14,9 +14,10 @@ import pastis.util as util
 if __name__ == '__main__':
 
     APLC_DESIGN = 'small'
-
+    poke_amplitude = 1e-8 #m
     # First generate a couple of matrices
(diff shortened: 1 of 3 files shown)


ivalaginja:
Hey @ananyasahoo1904, adding the calibration aberration as input parameter to the matrix generation is a great idea, however, I would like to talk about where the input variable gets defined. The reason I was keeping it in the configfile so far was to keep things traceable, since the configfile gets saved out with every new matrix.
The only small difference this would make for your PR is to still keep the definition of `a_c` in the configfile, and simply read it from there in the launcher, and then pass it to the refactored matrix generation classes.

asahooexo:
This seemed to be a multiline change in the main script. As it is not the main priority and due to time constraint, closing this draft.

PR #124, opened 2022-06-19
Create class for calculation of sensitivity matrix of HexRing telescopes

ivalaginja:
This PR adds the required class and function to calculate a sensitivity matrix (in E-fields) for the new HexRing telescopes with an APLC. fyi @brynickson this is another PR for your SPIE proceedings

@ananyasahoo1904 this class only calculates the sensitivity matrix in the sicence focal plane, but it should be easy to adapt it to also generate a sensitivity matrix in the wavefront sensor plane.


ivalaginja:
Ready for testing... will need to test each touched class and function offline since they are not covered by the pytests.

ivalaginja:
Waiting for another bugfix in #127.

ivalaginja:
The script runs but the real parts of the focal-plane modes seem scrambled when run for the 1-Hex aperture. They are fine for LUVOIR-A. I will have to investigate before moving on.

asahooexo:
@ivalaginja why can't we merge them (focal plane + wavefront sensor plane) under single class?

If we divide them separately, this PR essentially means calculating pastis matrix (which is already in the repo)
As far as I recall you strongly wanted to group all G matrices and PASTIS matrix under one class.

ivalaginja:
@ananyasahoo1904 calculating both matrices in the same process/class - yes that is still the plan. But it doesn't mean that both implementations need to be made in the same PR. On the contrary, it is better to chunk up the work in separate PRs, and make self-consistent changes along the way that also get tested in each step.

The scope of this PR here is to calculate the sensitivity matrix for the Hex apertures in the science plane, as we were doing it before. The next batch of work, and hence the next PR, can integrate the calculation of the sensitivity matrix in the WFS plane - in the same class though.

ivalaginja:
This should be good to go now. Both the LUVOIR as well as 1-Hex matrix launcher work.

LUVOIR-A sensitivity matrix for 3 local Zernikes on each segment:
Screen Shot 2022-06-22 at 10 15 35

1-Hex sensitivity matrix for 3 local Zernikes on all 7 segments:
Screen Shot 2022-06-22 at 10 17 18

PR #125, opened 2022-06-19
Simplify matrix calculation classes

ivalaginja:
Some general changes to the top-level classes for sensitivity matrix calculation.

- removed `design` from input parameters to `PastisMatrix`, this attribute is now set only in the Luvoir classes
- remove inheritance from `ABC`
- rename output path variable in `PastisMatrix`
- change variable `nb_seg` such that it is now passed into `PastisMatrix` upon instantiation instead of reading it from the configfile directly there.


ivalaginja:
Waiting on #126 to be merged.

PR #128, opened 2022-06-21
Create functions for calculation of sensitivity matrices at the wfs plane

asahooexo:
This PR adds function to calculate electric field (can be ideal_reference or poked efields) at the wavefront sensor plane.  These fields are further used to generate sensitivity matrices at the wavefront sensor plane. 

1. Add calculate_ref_efield_wfs(): This function generates an electric field as seen by the wavefront sensor when there is no aberration at the pupil plane.
2. Add calculate_efields_wfs(): This function generates a list of electric fields at the wavefront sensor plane by poking the deformable mirror with individual aberration modes
3. save the real and imaginary part of the efields_wfs
4. add norm_one_photon parameter to the all the efields/matrices calculated. 
5. norm_one_photon to be kept same across all efields/matrices computation

### DIFF
--- a/pastis/matrix_generation/matrix_from_efields.py
+++ b/pastis/matrix_generation/matrix_from_efields.py
@@ -43,6 +43,7 @@ class PastisMatrixEfields(PastisMatrix):
         self.saveopds = saveopds
         self.calculate_one_mode = None
         self.efields_per_mode = []
+        self.efields_per_mode_wfs = []
 
         os.makedirs(os.path.join(self.resDir, 'efields'), exist_ok=True)
 
@@ -68,6 +69,13 @@ class PastisMatrixEfields(PastisMatrix):
         for i in range(self.number_all_modes):
             self.efields_per_mode.append(self.calculate_one_mode(i))
         self.efields_per_mode = np.array(self.efields_per_mode)
+
+    def calculate_efields_wfs(self):
+        """ Poke each mode individually and calculate the resulting focal plane E-field. """
+
+        for i in range(self.number_all_modes):
+            self.efields_per_mode_wfs.append(self.calculate_one_mode(i)) #TODO: does this function needs to be changed?
+        self.efields_per_mode_wfs = np.array(self.efields_per_mode_wfs)
 
     def calculate_pastis_matrix_from_efields(self):
         """ Use the individual-mode E-fields to calculate the PASTIS matrix from it. """
@@ -195,6 +203,15 @@ class MatrixEfieldInternalSimulator(PastisMatrixEfields):
         # Calculate reference E-field in focal plane, without any aberrations applied
         unaberrated_ref_efield, _inter = self.simulator.calc_psf(return_intermediate='efield')
         self.efield_ref = unaberrated_ref_efield.electric_field
+
+    def calculate_ref_efield_wfs(self, norm_one_photon=False):
+        """Calculate the reference E-field at the wavefront sensor plane."""
+        if norm_one_photon:
+            unaberrated_ref_efield_wfs = self.simulator.calc_out_of_band_wfs(norm_one_photon=True)
+        else:
+            unaberrated_ref_efield_wfs = self.simulator.calc_out_of_band_wfs(norm_one_photon=False)
+
+        self.efield_ref_wfs = unaberrated_ref_efield_wfs
 
     def setup_deformable_mirror(self):
         """ Set up the deformable mirror for the modes you're using and define the total number of mode actuators. """
--- a/pastis/matrix_generation/matrix_from_efields.py
+++ b/pastis/matrix_generation/matrix_from_efields.py
@@ -260,8 +260,8 @@ class MatrixEfieldInternalSimulator(PastisMatrixEfields):
         """ Create the partial function that returns the E-field of a single aberrated mode. """
 
         self.calculate_one_mode = functools.partial(_simulator_matrix_single_mode, self.which_dm, self.number_all_modes,
-                                                    self.wfe_aber, self.simulator, self.norm_one_photon, self.resDir, self.save_efields,
-                                                    self.saveopds)
+                                                    self.wfe_aber, self.simulator, self.calc_science, self.calc_wfs,
+                                                    self.norm_one_photon, self.resDir, self.save_efields, self.saveopds)
 
 
 class MatrixEfieldLuvoirA(MatrixEfieldInternalSimulator):
@@ -368,21 +368,27 @@ class MatrixEfieldRST(PastisMatrixEfields):
                                                     self.rst_cgi, self.resDir, self.save_efields, self.saveopds)
 
 
-def _simulator_matrix_single_mode(which_dm, number_all_modes, wfe_aber, simulator, norm_one_photon, resDir, saveefields, saveopds, mode_no):
+def _simulator_matrix_single_mode(which_dm, number_all_modes, wfe_aber, simulator, calc_science, calc_wfs,
+                                  norm_one_photon, resDir, saveefields, saveopds, mode_no):
     """
     Calculate the mean E-field of one aberrated mode on one of the internal simulator instances; for PastisMatrixEfields().
     :param which_dm: string, which DM - "seg_mirror", "harris_seg_mirror", "zernike_mirror"
     :param number_all_modes: int, total number of all modes
     :param wfe_aber: float, calibration aberration in meters
     :param simulator: instance of segmented telescope simulator
+    :param calc_science: bool, whether to calculate the Efields in the science focal plane.
+    :param calc_wfs: bool, whether to calculate the Efields in the out-of-band Zernike WFS plane.
+    :param norm_one_photon: bool, whether to normalize the returned E-fields and intensities to one photon in the entrance pupil.
     :param resDir: str, directory for matrix calculation results
     :param saveefields: bool, Whether to save E-fields as fits file to disk or not
     :param saveopds: bool, Whether to save images of pair-wise aberrated pupils to disk or not
     :param mode_no: int, which mode index to calculate the E-field for
-    :return: complex array, resulting focal plane E-field
+    :return: dict, resulting E-fields
     """
 
     log.info(f'MODE NUMBER: {mode_no}')
+    efield_focal_plane = None
+    efield_wfs_plane = None
 
     # Apply calibration aberration to used mode
     all_modes = np.zeros(number_all_modes)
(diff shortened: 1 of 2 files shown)


ivalaginja:
Looks good so far! @ananyasahoo1904 could you rebase please, since I merged #124 in the meantime? Or, since there are merge conflicts, let me know if you would prefer that I do the rebase.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -66,9 +71,15 @@ def calculate_efields(self):
         """ Poke each mode individually and calculate the resulting focal plane E-field. """
 
         for i in range(self.number_all_modes):
-            self.efields_per_mode.append(self.calculate_one_mode(i))
+            self.efields_per_mode.append(self.calculate_one_mode(i)[0])
         self.efields_per_mode = np.array(self.efields_per_mode)
 
+    def calculate_efields_wfs(self):
+        """ Poke each mode individually and calculate the resulting focal plane E-field. """
```
ivalaginja: Please add in this docstring that this is in the WFS plane.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -166,12 +181,12 @@ def instantiate_simulator(self):
         """ Create a simulator object and save to self.simulator """
         raise NotImplementedError()
 
-    def calculate_ref_efield(self):
+    def calculate_ref_efield(self, norm_one_photon):
```
ivalaginja: You are passing a variable `norm_one_photon` to this method, but in the end, in line 189, you actually use the class attribute `self.norm_one_photon` instead of the variable, which means that the variable remains unused. Considering that the method `calculate_ref_efield()` is a class method required by `PastisMatrixEfields`, which is also used by RST, which does not have the option to normalize its outputs, I would remove the input variable to the method and keep using the class attribute. This would mean:
- [x] remove the variable in the function signature here (line 184)
- [x] remove the variable in the function signature for `calculate_ref_efield_wfs()` in line 210
- [x] remove the variable in the function signature in lines 94 and 98
- [x] remove the variable in the function calls in lines 57 and 58

In this way, you don't need to adjust this method inside of `MatrixEfieldRST` and can remove the question in the todo comment there.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -276,14 +296,14 @@ def __init__(self, which_dm, dm_spec, num_rings=1, initial_path='', saveefields=
...
-                         saveefields=saveefields, saveopds=saveopds)
+                         saveefields=saveefields, saveopds=saveopds, norm_one_photon=norm_one_photon)
 
     def instantiate_simulator(self):
         optics_input = os.path.join(util.find_repo_location(), 'data', 'SCDA')
         sampling = CONFIG_PASTIS.getfloat('HexRingTelescope', 'sampling')
         self.simulator = HexRingAPLC(optics_input, self.num_rings, sampling)
 
-
+#TODO: add norm_one_photon?
```
ivalaginja: No, this is not needed if the comment further above about the function signature for `calculate_ref_efield()` and `calculate_ref_efield_wfs()` is followed, so this comment can be deleted.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -367,9 +397,9 @@ def _simulator_matrix_single_mode(which_dm, number_all_modes, wfe_aber, simulato
         hcipy.imshow_field(opd_map, grid=simulator.aperture.grid, mask=simulator.aperture, cmap='RdBu')
         plt.savefig(os.path.join(resDir, 'OTE_images', opd_name + '.pdf'))
 
-    return efield_focal_plane.electric_field
-
+    return efield_focal_plane.electric_field, efield_wfs_plane
 
+#TODO: add norm_one_photon?
```
ivalaginja: No, this is not needed, as this is a function that becomes a partial function inside the RST matrix class anyway, so it can have a completely different function call. Comment can be deleted.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -352,13 +372,23 @@ def _simulator_matrix_single_mode(which_dm, number_all_modes, wfe_aber, simulato
...
+
+    # Calculate WFS plane E-field
+    efield_wfs_plane = simulator.calc_out_of_band_wfs(norm_one_photon=norm_one_photon)
 
     if saveefields:
-        fname_real = f'efield_real_mode{mode_no}'
-        hcipy.write_fits(efield_focal_plane.real, os.path.join(resDir, 'efields', fname_real + '.fits'))
-        fname_imag = f'efield_imag_mode{mode_no}'
-        hcipy.write_fits(efield_focal_plane.imag, os.path.join(resDir, 'efields', fname_imag + '.fits'))
+        #Save focal plane Efields
```
ivalaginja: Please add a space after the `#`.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -352,13 +372,23 @@ def _simulator_matrix_single_mode(which_dm, number_all_modes, wfe_aber, simulato
...
+        fname_real_focal = f'focal_real_mode{mode_no}'
+        hcipy.write_fits(efield_focal_plane.real, os.path.join(resDir, 'efields', fname_real_focal + '.fits'))
+        fname_imag_focal = f'focal_imag_mode{mode_no}'
+        hcipy.write_fits(efield_focal_plane.imag, os.path.join(resDir, 'efields', fname_imag_focal + '.fits'))
+
+        #Save wfs plane Efields
```
ivalaginja: Please add a space after the `#`.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -367,9 +397,9 @@ def _simulator_matrix_single_mode(which_dm, number_all_modes, wfe_aber, simulato
         hcipy.imshow_field(opd_map, grid=simulator.aperture.grid, mask=simulator.aperture, cmap='RdBu')
         plt.savefig(os.path.join(resDir, 'OTE_images', opd_name + '.pdf'))
 
-    return efield_focal_plane.electric_field
-
+    return efield_focal_plane.electric_field, efield_wfs_plane
```
ivalaginja: This change is not trivial, as it would also require the function `_rst_matrix_single_mode()` to return two objects rather than just one. Let's talk about how we can solve this.

In any case, if we chose to keep this change, the E-field in the WFS plane would very likely also need to return the actual efield attribute `.electric_field`, like it is done for the focal-plane E-field.

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -66,9 +71,15 @@ def calculate_efields(self):
         """ Poke each mode individually and calculate the resulting focal plane E-field. """
 
         for i in range(self.number_all_modes):
-            self.efields_per_mode.append(self.calculate_one_mode(i))
+            self.efields_per_mode.append(self.calculate_one_mode(i)[0])
```
ivalaginja: I would like to revisit this change separately, see comment in line 400. (1/2)

pastis/matrix_generation/matrix_from_efields.py
```diff
@@ -66,9 +71,15 @@ def calculate_efields(self):
...
         self.efields_per_mode = np.array(self.efields_per_mode)
 
+    def calculate_efields_wfs(self):
+        """ Poke each mode individually and calculate the resulting focal plane E-field. """
+        for i in range(self.number_all_modes):
+            self.efields_per_mode_wfs.append(self.calculate_one_mode(i)[1])
```
ivalaginja: I would like to revisit this change separately, see comment in line 400. (2/2)

ivalaginja (changes requested):
Very nice implementation. This is doing what it needs to do, although we will have to make sure to test this offline and post some figures of the results.

One big comment is that while this calculates the E-fields in the WFS plane, it does not actually calculate the sensitivity matrix for that plane. Since that would just mean writing a new function and calling it in the `calc()` method of `PastisMatrixEfields`, we should include that in this PR.

Of the comments I left, please address the ones that you can right now and let me know when you are done, then we can talk about how to solve the issue with the `self.calculate_one_mode()` method. I currently have no good ideas for a solution, but I will think about it.

asahooexo:
Calculating sensitivity matrices (G) literally just means stacking efields_wfs, Sometimes we first subsample the fields and then stack them to make G matrices. I wanted the subsample factor to be a user defined. Therefore, didn't add here to make it more complicated.

ivalaginja:
Does the subsampling need to be taken into account in the E-field measurement already or can that be done directly when creating the sensitivity matrix?

asahooexo:
While creating PASTIS matrices, subsampling wasn't taken into account. 
We create two matrices, one is G_coron(efields from pastis) and G_OBWFS (efields from wfs). Currently G_OBWFS is only subsampled. I remembered Laurent saying it was done to run it faster. (I'll recheck again)

I don't think you need subsampling factor while calculating efields. We use hcipy.field.subsample_field() after efields have been generated

ivalaginja:
Ok. In that case it should be easy to make a `subsample_wfs` input parameter in the same way you added `norm_one_photon`. Otherwise we'd need a completely separate function to calculate that matrix, which I think is not necessary. Or do you anticipate to keep recalculating matrices once the E-fields are measured?

asahooexo:
Yes, I anticipate recalculating G matrices with different subsampling ratio. I feel like the subsampling ratio is dependent on simulator's grid. I want to play with this factor separately.

asahooexo:
currently, 
the subsampling is done as follows:
```
N_pup_z = int(simulator.pupil_grid.shape[0] / z_pup_downsample)
grid = hcipy.field.make_pupil_grid(N_pup_z, diameter=simulator.diam)
```
```
efield = simulator.calc_out_of_band_wfs(norm_one_photon=True)
efield_sub_real = hcipy.field.subsample_field(efield.real, z_pup_downsample, grid, statistic='mean')
```

This code was breaking for HeX, I still have to debug

asahooexo:
I think I have addressed all your immediate changes, except `calculate_one_mode`.
I think this PR is ready to be merged. 

In the next PR, I plan to put G_OBWFS, and G_coron

asahooexo:
Also, I tested this script offline for luvoirA, norm_one_photon=True/False. The entire matrix generation takes 30 mins on my local machine.

asahooexo:
Is there anything left to be done from my end?

ivalaginja:
This looks good 👍 

I added three changes:
- Rewrite the WFS Efield propagation so that we don't repeat the calculation of all the optical planes, just add the extra propagation to the WFS.
- Reformat the return of the functions `_simulator_matrix_single_mode()` and `_rst_matrix_single_mode()` so that they now return a dictionary of E-fields. In this way, if not all such functions return the same E-field, it is fine.
- Add the boolean input parameters `calc_science` and `calc_wfs` to be able to switch the respective E-field calculations on and off.

ivalaginja (approved):
Good to go!

PR #129, opened 2022-06-24
Add full suite of SCDA simulators

ivalaginja:
Also for Nickson et al. SPIE 2022 @brynickson.
fyi @ananyasahoo1904

This PR finally also adds a bunch of tests for the internal simulator, at least a subset.

- [x] fix segment flat to flat sizes for 3-,4-,5-Hex designs
- [x] add necessary files to repo


ivalaginja:
I now updated the segment flat-to-flat diameter values with the results I calculated with the equation provided by @brynickson. I am still not sure why those were off and I hope we can follow up on that.

For now, just the tests need to pass and this is ready to be merged.

PR #134, opened 2022-07-11
Read operating wavelength from configfile

ivalaginja:
I had put the correct wavelength into the configfile but never made the SCDA simulator read it from there 😬 
This is fixed now.

fyi @ananyasahoo1904 @brynickson


ivalaginja:
The failed tests are kinda bogus tests that I will fix separately. Considering everything else passed, I am merging this.

PR #136, opened 2022-07-18
Fix poor contrast performance of Luvex simulator

ivalaginja:
The 5 SCDA telescopes designs with their APLCs were all designed to reach an average DH mean contrast below 1e-10. We noticed that this performance was not reached in the current implementation of the simulator, except for the 1-Hex and the 2-Hex design.

The reasons for this are:
- The circumscribed diameters encoded in the fits file headers of the SCDA telescope designs were scrambled and not entirely correct. @brynickson remeasured them in DS9 under the assumption that the *inscribed* diameter is exactly 6m for all five designs, and calculated the *circumscribed* (point to point) diameters from that, resulting in the following five numbers for the five designs: `[7.952780833, 7.258823341, 7.72593506, 7.136856303, 6.844452477]`
- The Lyot stops, which all have a relative size of 98.2% of the respective inscribed diameter, also had slightly scrambled sizes.

This PR provides a temporary fix for this issue although the goal is to come back to this and understand what really caused this. The fixes that are implemented here are:
- Instead of reading the circumscribed diameters of the apertures from the fits files, they are now taken from the above list.
- The Lyot stops for the 3-, 4- and 5-Hex designs are now created with HCIPy on the same pupil grid like the apertures get sampled on, with diameters that were found by means of a parameter sweep, where we picked the value that gave the best contrast.

The problem that remains is that the DH morphologies are still slightly off from the design process of the APLC designs. The only design that is perfectly fine (by reading both the aperture and the LS from file) is the 2-Hex design - both the contrast performance as well as the DH morphology are what we expect from the design process. Meanwhile, the 1-hex design meets the contrast spec but does not match the DH morphology out of the box. Both of these designs are not touched in this PR.

Below a comparison between the expected DHs and DHs produced by the fixed simulator in this PR:

1-Hex expected:
Screen Shot 2022-07-18 at 15 57 57

1-Hex current:
Screen Shot 2022-07-18 at 15 58 43

2-Hex expected:
Screen Shot 2022-07-18 at 15 59 07

2-Hex current:
Screen Shot 2022-07-18 at 15 59 36

3-Hex expected (left) and current (right):
Screen Shot 2022-07-18 at 16 00 52

4-Hex expected (left) and current (right):
Screen Shot 2022-07-18 at 16 01 00

5-Hex expected (left) and current (right):
Screen Shot 2022-07-18 at 16 01 12

Incidentally, all the changes in this PR keep the segment size in the modal basis used for the segmented aberrations correct; there are potentially a handful of blank pixels now showing up on the edges of the segment modes, but this should not impact the performance of the simulator.

### DIFF
--- a/notebooks/SCDA/1_test-scda-simulator.ipynb
+++ b/notebooks/SCDA/1_test-scda-simulator.ipynb
@@ -34,8 +34,8 @@
     "# Define input parameters\n",
     "optics_dir = os.path.join(find_repo_location(), 'data', 'SCDA')\n",
     "sampling = 4\n",
-    "num_rings = 1\n",
-    "robust = 4"
+    "num_rings = 3\n",
+    "robust = 0"
    ]
   },
   {
@@ -89,9 +89,9 @@
    "outputs": [],
    "source": [
     "tel.flatten()\n",
-    "tel.set_segment(2, 1e-4, 0, 0)\n",
-    "tel.set_segment(4, 0, 1e-4, 0)\n",
-    "tel.set_segment(7, 0, 0, 1e-4)"
+    "tel.set_segment(1, 1e-8, 0, 0)\n",
+    "tel.set_segment(4, 0, 1e-8, 0)\n",
+    "tel.set_segment(7, 0, 0, 1e-8)"
    ]
   },
   {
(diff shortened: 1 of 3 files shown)


ivalaginja:
The DH average contrasts measured in the current DHs are:
Screen Shot 2022-07-18 at 15 18 26

PR #138, opened 2022-09-01
Add analysis scripts for scda telescope

asahooexo:
This PR needs to be merged after PR #130. 
This adds the following features to the existing repo.

- [x] Create PASTIS matrix for scda telescopes with segment level multimode mode aberration.
- [ ] Generate static tolerance plots and maps
- [ ] mcmc simulation to validate the static tolerances
- [ ] call the focal plane and wfs plane efields generated while pastis matrix, and reshape them to feed them into temporal analysis funcs
- [ ] generate contrast vs t_wfs curves
- [ ] and use the optimal t_wfs found from the curve to fill error budget table.
- [ ] validate the temporal tolerance quantites using close loop function

Currently, most of this step is calculated only in a single big scrips. Therefore, breaking this PR into several mini PRs.

### DIFF
--- a/pastis/temporal_analysis/run_luvoir_analysis.py
+++ b/pastis/temporal_analysis/run_luvoir_analysis.py
@@ -340,3 +340,9 @@ c5 = c_contrast5['averaged_hist']
 n_tmp1 = len(c5)
 result_c5.append(c5[n_tmp1-1])
 print(result_c5[0]-contrast_floor)
+
+print(np.sqrt(np.mean((result_c5[0]-contrast_floor)**2 +
+                      (result_c4[0]-contrast_floor)**2 +
+                      (result_c3[0]-contrast_floor)**2 +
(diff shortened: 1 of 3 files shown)


pastis/temporal_analysis/run_luvoir_analysis.py
```diff
@@ -0,0 +1,224 @@
...
+def matrix_subsample_fast(matrix, n, m):
+    l = matrix.shape[0] // n  # block length
+    b = matrix.shape[1] // m  # block breadth
+    new_shape = (n, l, m, b)
+    reshaped_array = matrix.reshape(new_shape)
+    data_reduced = np.sum(reshaped_array, axis=(1, 3))
+    return data_reduced
+
+
+def calculate_sensitivity_matrices(e0_coron, e0_obwfs, efield_coron_real,
```
asahooexo, line 38: @ivalaginja here is where the sensitivity matrices are calculated.

PR #139, opened 2022-09-01
Write out files to temporary directory.

ehpor:
This PR changes the output path of the pytests from a local directory to a temporary directory. This avoids clogging up the current working directory when running the pytests locally.

I'm not sure if writing to the CWD is intentional or not, since it might help with diagnosis in some cases. Feel free to close this PR if this is the case.

Tests will fail until #137 is merged, so I'm keeping this in draft until that happened and this is rebased.


ivalaginja (approved):
Thanks for this @ehpor !

PR #141, opened 2022-10-14
Add multimode static tolerance plot.

asahooexo:
This PR generates multimode static tolerance plots for the luvex and luvoir simulator.

I added additional lines to the launcher scripts, plotting.py and util.py, which computes static tolerance maps for Harris thermal modes as well localized zernike modes.

For Segment level aberrations such as Zernike/harris modes, it creates another directory named "mu_maps" inside the parent run where the pdfs of these maps are stored. 

Following is an example of 1-ring-luvex simulator, where 10 tolerance modes are saved for first 10 segment-level zernike modes. Make sure you instantiate the simulator with exact number modes, as you use before plotting

Screen Shot 2022-10-24 at 6 08 30 PM

Below is a sample picture of one segment level-zernike mode:
Screen Shot 2022-10-24 at 6 10 52 PM

For harris mode:
here is sample of one tolerance map for 5-ring-luvex simulator:

Screen Shot 2022-10-24 at 6 36 20 PM

Also, it plots tolerance values per seg per zernike mode

Screen Shot 2022-10-24 at 7 34 08 PM

Also for harris mode:
Screen Shot 2022-10-24 at 7 36 31 PM

### DIFF
--- a/pastis/launchers/run_luvoir_efields.py
+++ b/pastis/launchers/run_luvoir_efields.py
@@ -1,12 +1,13 @@
 import os
 from astropy.io import fits
 import numpy as np
+
 from pastis.config import CONFIG_PASTIS
 import pastis.util as util
 from pastis.matrix_generation.matrix_from_efields import MatrixEfieldLuvoirA
-from pastis.pastis_analysis import  calculate_segment_constraints
-from pastis.simulators.luvoir_imaging import LuvoirA_APLC
+from pastis.pastis_analysis import calculate_segment_constraints
 import pastis.plotting as ppl
+
 
 if __name__ == '__main__':
 
--- a/pastis/launchers/run_luvoir_efields.py
+++ b/pastis/launchers/run_luvoir_efields.py
@@ -55,4 +55,4 @@ if __name__ == '__main__':
 
     tel = run_matrix.simulator
     ppl.plot_multimode_mus_surface_map(tel, mus, num_modes, tel.sm.num_actuators,
-                                       c_target, dir_run, mirror='sm', save=False)
+                                       c_target, dir_run, mirror='sm', cmin=-5, cmax=5, save=True)
--- a/pastis/launchers/run_luvoir_efields.py
+++ b/pastis/launchers/run_luvoir_efields.py
@@ -54,5 +54,4 @@ if __name__ == '__main__':
     ppl.plot_thermal_mus(mus, num_modes, nseg, c_target, out_dir=dir_run, save=True)
 
     tel = run_matrix.simulator
-    ppl.plot_multimode_mus_surface_map(tel, mus, num_modes, tel.sm.num_actuators,
-                                       c_target, dir_run, mirror='sm', cmin=-5, cmax=5, save=True)
+    ppl.plot_multimode_mus_surface_map(tel, mus, num_modes, c_target, dir_run, mirror='sm', cmin=-5, cmax=5, save=True)
--- a/pastis/launchers/run_luvoir_efields.py
+++ b/pastis/launchers/run_luvoir_efields.py
@@ -9,25 +9,29 @@ from pastis.simulators.luvoir_imaging import LuvoirA_APLC
 import pastis.plotting as ppl
 
 if __name__ == '__main__':
+
+    NUM_RINGS = 1
+    WHICH_DM = 'harris_seg_mirror'   # 'harris_seg_mirror' or 'seg_mirror', or (global) 'zernike_mirror'
+
     # DM_SPEC = tuple or int, specification for the used DM -
     # for seg_mirror: int, number of local Zernike modes on each segment
     # for harris_seg_mirror: tuple (string, array, bool, bool, bool),
     # absolute path to Harris spreadsheet, pad orientations, choice of Harris mode sets (thermal, mechanical, other)
     # for zernike_mirror: int, number of global Zernikes
 
-    # If Harris deformable mirror, uncomment the following lines
-    # DM = 'harris_seg_mirror'  # Possible: "seg_mirror", "harris_seg_mirror", "zernike_mirror"
-    # fpath = CONFIG_PASTIS.get('LUVOIR', 'harris_data_path')  # path to Harris spreadsheet
-    # pad_orientations = np.pi / 2 * np.ones(CONFIG_PASTIS.getint('LUVOIR', 'nb_subapertures'))
-    # DM_SPEC = (fpath, pad_orientations, True, False, False)
+    # If using Harris deformable mirror
+    if WHICH_DM == 'harris_seg_mirror':
+        fpath = CONFIG_PASTIS.get('LUVOIR', 'harris_data_path')  # path to Harris spreadsheet
+        pad_orientations = np.pi / 2 * np.ones(CONFIG_PASTIS.getint('LUVOIR', 'nb_subapertures'))
+        DM_SPEC = (fpath, pad_orientations, True, False, False)
 
-    # If Segmented Zernike Mirror, uncomment the following lines
-    DM = 'seg_mirror'  # Possible: "seg_mirror", "harris_seg_mirror", "zernike_mirror"
-    DM_SPEC = 5
+    # If using Segmented Zernike Mirror
+    if WHICH_DM == 'seg_mirror':
+        DM_SPEC = 3
 
     APLC_DESIGN = 'small'
     # First generate a couple of matrices
-    run_matrix = MatrixEfieldLuvoirA(which_dm=DM, dm_spec=DM_SPEC, design=APLC_DESIGN,
+    run_matrix = MatrixEfieldLuvoirA(which_dm=WHICH_DM, dm_spec=DM_SPEC, design=APLC_DESIGN,
                                      calc_science=True, calc_wfs=True,
                                      initial_path=CONFIG_PASTIS.get('local', 'local_data_path'),
                                      norm_one_photon=True)
@@ -35,7 +39,7 @@ if __name__ == '__main__':
     dir_run = run_matrix.overall_dir
     print(f'All saved to {dir_run}.')
 
-    # Calculate and save per segment static tolerance
+    # get the automatically saved pastis_matrix
     pastis_matrix = fits.getdata(os.path.join(dir_run, 'matrix_numerical', 'pastis_matrix.fits'))
 
     # get the unaberrated coro_psf after the matrix run
@@ -43,15 +47,24 @@ if __name__ == '__main__':
     dh_mask = np.array(run_matrix.simulator.dh_mask.shaped)
     contrast_floor = util.dh_mean(e0_psf, dh_mask)
 
-    # set the target contrast
+    # Calculate the static tolerances
     c_target = 5.3*1e-11
     mus = calculate_segment_constraints(pastis_matrix, c_target=c_target, coronagraph_floor=contrast_floor)
-    np.savetxt(os.path.join(dir_run, 'mu_map_%s.csv' % c_target), mus, delimiter=',')
+    np.savetxt(os.path.join(dir_run, f'mu_map_{c_target:.2e}.csv'), mus, delimiter=',')
 
     num_modes = 5  # for harris thermal map or number of localized zernike modes
     nseg = run_matrix.simulator.nseg
 
-    ppl.plot_thermal_mus(mus, num_modes, nseg, c_target, out_dir=dir_run, save=True)
+    coeffs_table = util.sort_1d_mus_per_segment(mus, num_modes, nseg)
+    mu_list = []
+    label_list = []
+    for i in range(coeffs_table.shape[0]):
+        mu_list.append(coeffs_table[i])
+        if WHICH_DM == 'seg_mirror':
+            label_list.append(f'Zernike mode {i}')
+    if WHICH_DM == 'harris_seg_mirror':
+        label_list = ['Faceplates Silvered', 'Bulk', 'Gradient Radial', 'Gradient X lateral', 'Gradient Z axial']
 
+    ppl.plot_segment_weights(mu_list, dir_run, c_target, labels=label_list, fname=f'stat_1d_mus_{c_target:.2e}', save=True)
     tel = run_matrix.simulator
     ppl.plot_multimode_mus_surface_map(tel, mus, num_modes, c_target, dir_run, mirror='sm', cmin=-5, cmax=5, save=True)
--- a/pastis/launchers/run_luvoir_efields.py
+++ b/pastis/launchers/run_luvoir_efields.py
@@ -9,17 +9,6 @@ from pastis.simulators.luvoir_imaging import LuvoirA_APLC
 import pastis.plotting as ppl
 
 if __name__ == '__main__':
-
-    APLC_DESIGN = 'small'
-
-    # Instantiate the LUVOIR telescope
-    optics_input = os.path.join(util.find_repo_location(), CONFIG_PASTIS.get('LUVOIR', 'optics_path_in_repo'))
-    coronagraph_design = CONFIG_PASTIS.get('LUVOIR', 'coronagraph_design')
-    sampling = CONFIG_PASTIS.getfloat('LUVOIR', 'sampling')
-    tel = LuvoirA_APLC(optics_input, coronagraph_design, sampling)
-    nm_aber = CONFIG_PASTIS.getfloat('LUVOIR', 'calibration_aberration') * 1e-9
-
-    # -----------------------------
     # DM_SPEC = tuple or int, specification for the used DM -
     # for seg_mirror: int, number of local Zernike modes on each segment
     # for harris_seg_mirror: tuple (string, array, bool, bool, bool),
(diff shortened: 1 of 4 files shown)


pastis/launchers/run_luvoir_efields.py
```diff
@@ -12,15 +17,51 @@
...
+    # Instantiate the LUVOIR telescope
+    optics_input = os.path.join(util.find_repo_location(), CONFIG_PASTIS.get('LUVOIR', 'optics_path_in_repo'))
+    coronagraph_design = CONFIG_PASTIS.get('LUVOIR', 'coronagraph_design')
+    sampling = CONFIG_PASTIS.getfloat('LUVOIR', 'sampling')
+    tel = LuvoirA_APLC(optics_input, coronagraph_design, sampling)
+    nm_aber = CONFIG_PASTIS.getfloat('LUVOIR', 'calibration_aberration') * 1e-9
+
+    # Create harris deformable mirror
+    tel.create_segmented_harris_mirror(fpath, pad_orientations, thermal=True, mechanical=False, other=False)
+    tel.harris_sm
```
ivalaginja: Is there something missing here?

pastis/launchers/run_luvoir_efields.py
```diff
@@ -12,15 +17,51 @@
...
     run_matrix.calc()
     dir_run = run_matrix.overall_dir
     print(f'All saved to {dir_run}.')
+
+
+    # Instantiate the LUVOIR telescope
+    optics_input = os.path.join(util.find_repo_location(), CONFIG_PASTIS.get('LUVOIR', 'optics_path_in_repo'))
+    coronagraph_design = CONFIG_PASTIS.get('LUVOIR', 'coronagraph_design')
+    sampling = CONFIG_PASTIS.getfloat('LUVOIR', 'sampling')
+    tel = LuvoirA_APLC(optics_input, coronagraph_design, sampling)
```
ivalaginja: This is duplicate code, as you can assess the simulator that has been used in `MatrixEfieldLuvoirA` a couple of lines above with `MatrixEfieldLuvoirA.simulator`.

pastis/launchers/run_luvoir_efields.py
```diff
@@ -12,15 +17,51 @@
...
+    num_actuators = tel.harris_sm.num_actuators
+    num_modes = 5
+
+    # calculate dark hole contrast
+    tel.harris_sm.flatten()
+    unaberrated_coro_psf, ref = tel.calc_psf(ref=True, display_intermediate=False, norm_one_photon=True)
+    norm = np.max(ref)
+    dh_intensity = (unaberrated_coro_psf / norm) * tel.dh_mask
+    contrast_floor = np.mean(dh_intensity[np.where(tel.dh_mask != 0)])
+    print(f'contrast floor: {contrast_floor}')
```
ivalaginja: If you run the analysis above, all of this has already been done and is either accessible within the class or saved to disk.

pastis/launchers/run_luvoir_efields.py
```diff
@@ -12,15 +17,51 @@
...
+
+    # Calculate and save per segment static tolerance
+    pastis_matrix = fits.getdata(os.path.join(dir_run, 'matrix_numerical', 'pastis_matrix.fits'))
+    c_target = 5.3*1e-11
+    mus = calculate_segment_constraints(pastis_matrix, c_target=c_target, coronagraph_floor=contrast_floor)
+    np.savetxt(os.path.join(dir_run, 'mu_map_harris_%s.csv' % c_target), mus, delimiter=',')
+
+    ppl.plot_thermal_mus(mus, num_modes, tel.nseg, c0=c_target, out_dir=dir_run, save=True)
+    ppl.plot_multimode_mus_surface_map(tel, mus, num_modes, num_actuators, c_target, dir_run, save=False)
```
ivalaginja: This would fit much better in an overall analysis script that we then call from the launcher. Otherwise the launcher gets more and more crowded depending on the types of analyses we want to do.

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+    Parameters
+    ----------
+    mus : ndarray
+        list of standard deviations for each segment
+    nmodes : int
+        number of thermal modes
+    nsegments :  int
+        number of segments
+    c0 : scalar
+        target contrast
```
ivalaginja: Let's make this We have been using `c_target` instead to stay consistent with the rest of the repo and also to make a distinction to the baseline contrast. Also, scalar is not a type, float is.

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+    out_dir : str
+        path to save the plot
```
ivalaginja: please add "if save=True"

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+    save : bool
+        whether to save the plot
```
ivalaginja: Please add a note on what happens if this is `False`

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+    """
+    harris_coeffs_table = np.zeros([nmodes, nsegments])
+    for qq in range(nmodes):
+        for kk in range(nsegments):
+            harris_coeffs_table[qq, kk] = mus[qq + (kk) * nmodes]
```
ivalaginja: Remove redundant prentheses.

Is this double-loop just rearanging the loaded data? If so, please make a little function out of it that we can reuse, since this is likely something we will be needing all over the place.

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+
+    plt.figure(figsize=(10, 10))
+    plt.title("Modal constraints to achieve a dark hole contrast of "r"$10^{%d}$" % np.log10(c0), fontsize=20)
```
ivalaginja: Please use f-strings for better readability.

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+    plt.plot(harris_coeffs_table[0]*1e3, label="Faceplates Silvered")
+    plt.plot(harris_coeffs_table[1]*1e3, label="Bulk")
+    plt.plot(harris_coeffs_table[2]*1e3, label="Gradiant Radial")
+    plt.plot(harris_coeffs_table[3]*1e3, label="Gradiant X lateral")
+    plt.plot(harris_coeffs_table[4]*1e3, label="Gradient Z axial")
+    plt.grid()
+    plt.legend(fontsize=15)
+    plt.tight_layout()
+    if save:
+        plt.savefig(os.path.join(out_dir, 'mus_1d_multi_modes_%s.png' % c0))
```
ivalaginja: f-string please

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+    """
+    harris_coeffs_table = np.zeros([nmodes, nsegments])
+    for qq in range(nmodes):
+        for kk in range(nsegments):
+            harris_coeffs_table[qq, kk] = mus[qq + (kk) * nmodes]
+
+    plt.figure(figsize=(10, 10))
+    plt.title("Modal constraints to achieve a dark hole contrast of "r"$10^{%d}$" % np.log10(c0), fontsize=20)
+    plt.ylabel("Weight per segment (in units of pm)", fontsize=15)
+    plt.xlabel("Segment Number", fontsize=20)
```
ivalaginja: Are we still doing segment numbers here or should they be mode numbers?

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+
+
+def plot_multimode_mus_surface_map(tel, mus, num_modes, num_actuators, c_target, data_dir, save=False):
+    """
+    Creates surface deformation tolerance maps for thermal aberrations.
+
+    Parameters:
+    tel : class instance of the simulator for "instrument"
+    mus : 1d array
+        list of tolerances for individual segments in units of nm
```
ivalaginja: Segments or modes? What is the required data structure of this array? Please add a comment to clarify.

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+    num_modes : int
+        len of segment level thmodal basis
```
ivalaginja: This description is unclear.

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+    num_actuators : int
+        total number of dm actuators
+    c_target : float
+        desired dark hole contrast for which the tolerancing is done
+    data_dir :  str
+        path to save the plot
+    save : bool
+        whether to save the plot
```
ivalaginja: Like above, please elaborate what happens in the case of save=False.

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+    """
+    nm_aber = CONFIG_PASTIS.getfloat('LUVOIR', 'calibration_aberration') * 1e-9
+
+    harris_coeffs_numaps = np.zeros([num_modes, num_actuators])
+    for qq in range(num_modes):
+        harris_coeffs_tmp = np.zeros([num_actuators])
+        for kk in range(tel.nseg):
+            harris_coeffs_tmp[qq + (kk) * num_modes] = mus[qq + (kk) * num_modes]  # arranged per modal basis
+        harris_coeffs_numaps[qq] = harris_coeffs_tmp  # arranged into 5 groups of nseg elements and in units of nm
```
ivalaginja: It seems like this would be a case where we reuse the function I asked you to write above.

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+
+    nu_maps = []
+    for qq in range(num_modes):
```
ivalaginja: Could you give this iteration variable a clearer name, e.g. `mode`?

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+    nm_aber = CONFIG_PASTIS.getfloat('LUVOIR', 'calibration_aberration') * 1e-9
+
+    harris_coeffs_numaps = np.zeros([num_modes, num_actuators])
+    for qq in range(num_modes):
+        harris_coeffs_tmp = np.zeros([num_actuators])
+        for kk in range(tel.nseg):
+            harris_coeffs_tmp[qq + (kk) * num_modes] = mus[qq + (kk) * num_modes]  # arranged per modal basis
+        harris_coeffs_numaps[qq] = harris_coeffs_tmp  # arranged into 5 groups of nseg elements and in units of nm
+
+    nu_maps = []
```
ivalaginja: Could we rename this to `mu_maps` to stay consistent with the mu we have been using all around?

pastis/plotting.py
```diff
@@ -994,3 +995,130 @@ def natural_keys(text):
...
+    plt.title("Gradient Z axial", fontsize=10)
+    plot_norm5 = TwoSlopeNorm(vcenter=0, vmin=-10, vmax=10)
+    hcipy.imshow_field((nu_maps[4]) * 1e12, norm=plot_norm5, cmap='RdBu')
+    plt.tick_params(top=False, bottom=False, left=False, right=False, labelleft=False, labelbottom=False)
+    cbar = plt.colorbar()
+    cbar.ax.tick_params(labelsize=10)
+    cbar.set_label("$pm$", fontsize=10)
+    plt.tight_layout()
+    if save:
+        plt.savefig(os.path.join(data_dir, 'stat_mu_maps_nm_%s.png' % c_target))
```
ivalaginja: f-string please

ivalaginja (changes requested):
Thanks for this work, I very much like the addition of the new plotting functions.

As for the part in the launcher, I would suggest to actually get started on a new analysis framework, rather than putting things into the launcher. I have left a couple of comments to highlight how this duplicates code, but I think this should not live there in the first place.

It should be possible to rewrite the existing analysis scripts to accommodate all the new modes we implemented, including the Harris modes. If this is not possible, it would be good to write a new, separate script that does the analysis in a systematic way. This can then be called from the launcher.

If you would like to get in touch about some ideas for this, please let me know.

pastis/launchers/run_luvoir_efields.py
asahooexo: @ivalaginja can you please point out how to access this?
I used util.dh_mean(im, dh) to get the contrast,
where im = fits.getdata(os.path.join(dir_run, 'unaberrated_coro_psf.fits'))
dh = run_matrix.simulator.dh_mask.shaped
And I get, contrast_floor = 5.39e-11

Whereas using the code I previously wrote, I get 4.636e-11

pastis/launchers/run_luvoir_efields.py
asahooexo: My mistake!!

pastis/launchers/run_luvoir_efields.py
asahooexo: I thought launchers to give a basic analysis. Okay, then where do you want to put these two plotting lines?

pastis/launchers/run_luvoir_efields.py
asahooexo: There seems to be a discrepancy between util.dh and my way of calculating contrast.

pastis/plotting.py
asahooexo: This means the total number of ways in which you can poke a segment. For a harris segment mirror  it is 5 (taking only thermal maps), for  segmented mirror, it is number of local zernike modes

pastis/plotting.py
asahooexo: Yes it is segment number. We want this very much.

pastis/launchers/run_luvoir_efields.py
asahooexo: This is due to the parameter "robutness_px". While independently instantiating the simulator I took robutness_px =4.0 and computed the contrast = 4.636e-11). But when run pastis matrix is calculated, automatically robutness_px is set to 0.0. 
For future errors, I should keep in mind the way I "manually "instantiate the simulator and the way simulator is instantiated while pastis matrix run. 
Therefore those two contrast values are for two different robustness_px. 
@ivalaginja should robustness_px be a hard coded parameter?

asahooexo:
@ivalaginja 
I have accommodated all your comments. Let me know if there is anything left. Hope this PR merges soon.

ivalaginja:
Hey @ananyasahoo1904 , could you post examples of the plots each of the new plotting function produces?

asahooexo:
@ivalaginja I have added screenshots of multiple plots this PR can generate.

asahooexo:
@ivalaginja all CIs have passed. Probably ready to merge?

pastis/plotting.py
```diff
@@ -994,3 +995,145 @@ def natural_keys(text):
...
+    """
+    nm_aber = CONFIG_PASTIS.getfloat('LUVOIR', 'calibration_aberration') * 1e-9
+
+    coeffs_mumaps = pastis.util.calculate_mu_maps(mus, num_modes, num_actuators, tel.nseg)
+
+    mu_maps = []
+    for qq in range(num_modes):
+        coeffs = coeffs_mumaps[qq] # in units of nm
+        if mirror == 'harris_sm':
+            tel.harris_sm.actuators = coeffs * nm_aber / 2 # in units of m
```
ivalaginja: @ananyasahoo1904 why do you multiply by `nm_aber` here and below when applying the coefficients to the mirror?

asahooexo: @ivalaginja The original mus generated from the PASTIS matrix were in units of poke aberration. Here poke aberration was 1nm, so units were of mus were 1 nm. I wanted to convert this to the SI units,  just to be on safer side and double checking.

pastis/plotting.py
ananyasahoo1904: you took care in R #142

asahooexo:
LGTM. @ivalaginja I crosse checked plot_segment_weights() works for single mode piston.

asahooexo:
@ivalaginja I addressed all your comments. Can it now be merged please?

pastis/plotting.py
ivalaginja: If `nm_aber` is 1 nm, multiplying by it doesn't do anything. All you are doing here is actually doing a unit conversion with the factor `1e-9` in the line where you read `nm_aber` from the configfile. If `nm_aber` were not 1 nm though, you would be scaling the actuator coefficients by an arbitrary number that was intended to calculate the matrix. The matrix itself is already normalized by this, see here:
https://github.com/spacetelescope/PASTIS/blob/0da795f43396d890fbf29f9d8e2072cdae66b69a/pastis/matrix_generation/matrix_from_efields.py#L159

This means that the `nm_aber` multiplication in the actuator command needs to be replaced just with the correct unit converstion to make it be meters, `1e-9` in this case here. I did this over in #143.

ivalaginja (approved):
Ready to go

PR #142, opened 2022-10-26
Delta-PR no.1 for multi-mode plotting

ivalaginja:
I suggest some changes in this PR that will make #141 slimmer while preserving all its new features.

In this PR, I:
- cleaned out and clarified the docstrings of all new functions
- simplified the new util functions that sort mode coefficients values in various ways
- removed showing of plots if not saved, otherwise there is no way to run the code silently. A `plt.show()` can always be set outside the plotting function itself if needed.
- consolidated `plot_zernike_mus()` with the already existing function `plot_segment_weights()` that does exactly that
- consolidated `plot_thermal_mus()` with the already existing function `plot_segment_weights()` that does exactly that

I also noticed that the thermal Harris mode map plotting does not work for Luvex launcher on the latest commit of #141 (9fcd41193700d53f320f4f683d5d6963dff2a663), or it is unclear how to set up the launcher for that. But we can work on this after we merge this PR into #141, and then I can do the second delta-PR, which will also implement a barebones analysis to be called from the launchers.

Note how this PR is not going into `develop` but into another branch, so there are no tests, just want your eyes on this before merging @ananyasahoo1904

PR #143, opened 2022-10-30
Delta-PR no.2 for multi-mode plotting

ivalaginja:
This is a follow-up to #142 and a delta-PR into #141.

In this PR:
- The multi-mode plotting functions is way more useful than just for tolerances, so I made it less specific to that by removing any reference to tolerance values. 
- I simplified the saving keyword to reduce one function variable, I intend to do this for other functions in the repo too.
- I clarified which of the input and output variables are in units of surface and which in units of WFE.
- I removed the `nm_aber` factor in the mode plotting as this was just an arbitrary scale factor. It happens to me one currently, but since this parameter is only used as an input to the matrix generation, it should not be used as a unit check in other places. See also https://github.com/spacetelescope/PASTIS/pull/141#discussion_r1008935479.
- I unified the file output names and file structure to what the repo has been doing for the analysis so far. I am willing to change this if there is a reason, otherwise I would like to keep this consistent to be able to run re-analyses without hickups.

I have not put the analysis in an external script yet, I will do that in a separate PR and after #141 is merged. This will be done in #144.

@asahooexo please have a look and test various telescope cases. I only had time to test for a 1-Hex telescope and that works fine, with the same outputs like before.

### DIFF
--- a/pastis/launchers/run_luvoir_efields.py
+++ b/pastis/launchers/run_luvoir_efields.py
@@ -67,4 +67,5 @@ if __name__ == '__main__':
 
     ppl.plot_segment_weights(mu_list, dir_run, c_target, labels=label_list, fname=f'stat_1d_mus_{c_target:.2e}', save=True)
     tel = run_matrix.simulator
-    ppl.plot_multimode_surface_maps(tel, mus, num_modes, c_target, dir_run, mirror='sm', cmin=-5, cmax=5, save=True)
+    os.makedirs(os.path.join(dir_run, 'mu_maps'), exist_ok=True)
+    ppl.plot_multimode_surface_maps(tel, mus, num_modes, mirror=WHICH_DM, cmin=-5, cmax=5, data_dir=dir_run, fname='stat_mu_maps')
--- a/pastis/launchers/run_luvoir_efields.py
+++ b/pastis/launchers/run_luvoir_efields.py
@@ -24,10 +24,12 @@ if __name__ == '__main__':
         fpath = CONFIG_PASTIS.get('LUVOIR', 'harris_data_path')  # path to Harris spreadsheet
         pad_orientations = np.pi / 2 * np.ones(CONFIG_PASTIS.getint('LUVOIR', 'nb_subapertures'))
         DM_SPEC = (fpath, pad_orientations, True, False, False)
+        NUM_MODES = 5  # TODO: works only for thermal modes currently
 
     # If using Segmented Zernike Mirror
     if WHICH_DM == 'seg_mirror':
         DM_SPEC = 3
+        NUM_MODES = DM_SPEC
 
     APLC_DESIGN = 'small'
     # First generate a couple of matrices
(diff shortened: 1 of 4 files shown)


pastis/launchers/run_luvoir_efields.py
```diff
@@ -24,10 +24,12 @@
         fpath = CONFIG_PASTIS.get('LUVOIR', 'harris_data_path')  # path to Harris spreadsheet
         pad_orientations = np.pi / 2 * np.ones(CONFIG_PASTIS.getint('LUVOIR', 'nb_subapertures'))
         DM_SPEC = (fpath, pad_orientations, True, False, False)
+        NUM_MODES = 5  # TODO: works only for thermal modes currently
 
     # If using Segmented Zernike Mirror
-    if WHICH_DM == 'seg_mirror':
+    if WHICH_DM in ['seg_mirror', 'zernike_mirror']:
         DM_SPEC = 3
+        NUM_MODES = DM_SPEC
```
asahooexo, line 32: I guess you added 'zernike_mirror' to do analysis for global zernike modes?

pastis/plotting.py
```diff
@@ -1000,57 +1000,58 @@ def natural_keys(text):
...
-            tel.harris_sm.actuators = coeffs * nm_aber / 2 # in units of m
-            mu_maps.append(tel.harris_sm.surface)
-        if mirror == 'sm':
-            tel.sm.actuators = coeffs * nm_aber / 2
-            mu_maps.append(tel.sm.surface)  # in units of m, each nu_map is now of the order of 1e-9 m
-
-    for i in range(0, num_modes):
+        coeffs = mus_per_actuator[mode]
+        if mirror == 'harris_seg_mirror':
+            tel.harris_sm.actuators = coeffs * 1e-9 / 2  # in meters of surface
```
asahooexo, line 1039: 1e-9 is a hardcoded parameter, I would still put nm_aber because this value used to poke and create PASTIS matrix in first place. If we change nm_aber to pm aberration in config file for calculating a PASTIS matrix, We will have to adjust these values accordingly.

From equation 17 and 36 in [Laginja et al. (2021)](https://ui.adsabs.harvard.edu/abs/2021JATIS...75004L/abstract), units of coeffs here is directly scaled to nm_aber.

[m_{ij}] = 1/{nm_aber^2}, [mu_k ] = nm_aber,  [] stands for dimension+scaling analysis, m_{ij} is and element of pastis matrix, mu_k is  wfe requirement per kth segment.

Also, I realized the units of pastis-matrix-elements is already normalized to 1 nm. The hardcoded parameter is  safe to write here. We might need to visit this later.

pastis/launchers/run_luvoir_efields.py
ivalaginja: Correct. I don't think it would work if we were to just run it blindly, but at least we remain the idea that this *should* be made possible further down the line.

asahooexo (commented):
I checked all the hex ring + mirror config config, It works fine. 
I would prefer to switch back to nm_aber and remove the hardcoded param. 

See my comment for L1039

pastis/plotting.py
ivalaginja: > From equation 17 and 36 in [Laginja et al. (2021)](https://ui.adsabs.harvard.edu/abs/2021JATIS...75004L/abstract), units of coeffs here is directly scaled to nm_aber.

Exactly! This is why the only conversion needed is from nm to m when feeding the coefficients into the simulator 👍

ivalaginja: See also:
https://github.com/spacetelescope/PASTIS/pull/141#discussion_r1008935479

PR #148, opened 2022-11-02
Add sensitivity function

asahooexo:
This PR stacks the electric fields at the wavefront sensor plane and focal plane saved during the parent PASTIS matrix run to create sensitvity matrices. These matrices are used to set temporal tolerances for the basis mode which was used in first place to create PASTIS matrix. 

Each of the efield is subtracted from its reference field at the same plane, and are stacked to be in a cube format. 

I subsample the electric field generated at the wavefront sensor plane for fast matrix algebra.


ivalaginja (commented):
I very much like the scope of this PR, it is compact and clear.

To proceed, I would like to ask you for the following changes:
- I think the new functions would live better in the module `pastis/temporal_analysis/close_loop_analysis.py` since it is there where we started collecting all functionality for the termporal analyses.
- Please replace all single-letter variables as they are cumbersome to work with (effectively un-searchable). This will also resolve the only flake8 error you currently have on this branch.
- Please add docstrings everywhere.
- Could you write a test or two that test the correct format of the resulting products? That way we make sure we don't break them by accident further down the line.

From my understanding, these are really important functions for your work on the analyses, so let's get them merged asap.

Issue #150, opened 2023-02-28
 unable to generate static multimode tolerance plots using plot_multimode_surface_maps in plotting.py 


asahooexo:
I tried running run_luvex_efields.py for 1 Hex configuration and called the plot_multimode_surface_maps in it. 
I get the following error in my machine. 

This is a path error. 

<img width="1601" alt="Screen Shot 2023-02-28 at 2 45 19 PM" src="https://user-images.githubusercontent.com/33359812/221962070-b962504e-dc84-4ed2-b34c-53bb686b6b2d.png">

Issue #140, opened 2022-09-02
Investigate regression test failure due to hcipy 0.5.0 upgrade


ivalaginja:
See #137

ivalaginja:
See also https://github.com/spacetelescope/PASTIS/pull/130#issuecomment-1227997114

ivalaginja:
Trying things to figure out what makes the tests fail with hcipy>0.4.0:

- It's not the aperture itself, when I create a Hex2 aperture with either hcipy 0.4.0 or 0.5.2, the resulting apertures used in the telescope object are identical and subtract perfectly to zero.
- The evaluated segments are also all identical (created with the method `_create_evaluated_segment_grid()`)
- The transformation matrix of the segment mode basis in the segmented deformable mirrors are the same between the two hcipy versions.

I found a difference in the produced efields. Here some differences between the efield in the final focal plane as calculated by `calc_psf()` of the telescope class - it's the imaginary part of the efield in this plane that changed significantly (bottom right plot). The top two plots are intensities, the bottom ones real and imaginary part of the efield. This was done with the normalization from the simulator, normalizing the total energy to one photon:
<img width="519" alt="Screenshot 2023-05-10 at 00 10 46" src="https://github.com/spacetelescope/PASTIS/assets/29508965/ea8aca71-f417-435e-b0fe-d4e266ce967a">

The difference is also visible in the focal plane of the FPM:
<img width="968" alt="Screenshot 2023-05-10 at 00 14 38" src="https://github.com/spacetelescope/PASTIS/assets/29508965/6bcb8cde-9f33-4405-b272-39a1d917b75a">

And in the Lyot plane:
<img width="960" alt="Screenshot 2023-05-10 at 00 16 26" src="https://github.com/spacetelescope/PASTIS/assets/29508965/8e0eea8e-3636-465d-8757-8b8d91c73ee5">

I repeated the data above without the normalization to make sure it's not that (and also centered the colorbar on zero):
![Screenshot 2023-05-10 at 10 54 51](https://github.com/spacetelescope/PASTIS/assets/29508965/41dfbfa6-77f2-4526-a22e-64122c6b54a5)
![Screenshot 2023-05-10 at 11 09 38](https://github.com/spacetelescope/PASTIS/assets/29508965/fe1f7fb5-004c-4443-b9c6-4d67f9133344)
[...]

PR #151, opened 2023-02-28
Fix path error in tolerance plots.

asahooexo:
This PR fixes a path error found while generating multimode tolerance plots. It creates a new folder mu_maps inside the parent directory where the data is saved.

Fixes issue #150.

### DIFF
--- a/pastis/plotting.py
+++ b/pastis/plotting.py
@@ -1223,5 +1223,5 @@ def plot_multimode_surface_maps(tel, mus, num_modes, mirror, cmin, cmax, data_di
 
         if data_dir is not None:
             os.makedirs(os.path.join(data_dir, 'mu_maps'), exist_ok=True)
-            fname += f'_mode_{i}.pdf'
-            plt.savefig(os.path.join(data_dir, 'mu_maps', fname))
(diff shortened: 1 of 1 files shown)


ivalaginja:
The tests are failing due to an issue in hcipy that was fixed very recently in https://github.com/ehpor/hcipy/pull/170, I will look into it.

ivalaginja (commented):
I am not sure this PR is needed in this way. The introduced change adds file directory handling to a function whose purpose is plotting. In my opinion, it's purpose should remain only plotting. The path to save the plot to is passed in and optional, in which case the responsibility of making sure the path exists lies in the function/script that calls the plotting function.

You found a nasty problem with a dependency though which we need to solve asap. This means I have to work on Issue #140, so I will put it on my priority list.

asahooexo:
I see. In this way, the user has to remember exactly the name of the directory to create before calling the 'plot_multimode_surface_maps'.

Also, the plot pdf names keeps on appending. This is how it looks like.

This PR also fixes this.
 
Screen Shot 2023-03-03 at 11 52 23 AM

Issue #152, opened 2023-03-13
Unpin scipy version


ivalaginja:
In #153, I pinned the scipy version due to a bug in hcipy that was fixed in https://github.com/ehpor/hcipy/pull/170 but is currently not released yet. Once that is done (hcipy release beyond 0.5.1) I can unpin scipy.

PR #153, opened 2023-03-13
Pin scipy version

ivalaginja:
Bug found by @asahooexo, same as in https://github.com/ehpor/hcipy/pull/170.

For now I will try to fix it by pinning the scipy version, but want to eventually fully fix the issue, see #152.


ivalaginja:
All tests are passing locally now

PR #154, opened 2023-03-14
Simplify pytests

### LINKED ISSUE #147: Reduce pytest runtime
The tests currently take so long because we are recalculating a bunch of stuff with the 120-segment LUVOIR simulator. Now that we have the Luvex simulators though, we can pick a design with less segments to speed up the tests.

ivalaginja:
- Fixes #147
- Sets calibration aberration for Hexring telescopes to 1 nm per single segment (see below)
- Pins matplotlib version to <=3.5 (see #155)

This PR gets rid of the full calculation of the 120-segment LUVOIR-A PASTIS matrix from intensities for a regression test. It is replaced with a PASTIS matrix calculation for the 2-Hex telescope, and from E-fields. This test now runs in a couple of minutes instead of 45 minutes to an hour.

I kept the LUVOIR matrix test files for now to be able to run the test that checks the calculation of the PASTIS matrix from a contrast matrix calculated form intensities. I might revisit this in the future, although the LUVOIR files are also needed for the PASTIS analysis tests.

---------

This PR also fixes the calibration aberration for the HexRing telescopes to 1 nm. A 1 nm piston WFE on a 19-segment aperture (2 rings) converts to a global WFE rms of 0.23 nm. (0.1 nm on a single segment yields a global WFE rms of 0.023 nm, 10 nm on one segment is 2.23 nm globally and 100 on one segment 22.94 nm globally.)

A 0.23 nm global rms WFE falls close to the knee of the hockey stick curve below, generated for a 2-Hex telescope with a sub 1e-10 APLC like on this repo. The x-axis shows the global WFE rms, the y-axis the normalized intensity in a circular DH from 3.4 to 12 $\lambda/D$.
Screenshot 2023-03-14 at 14 58 26

I looked at what the different calibration aberrations do to the matrix diagonal - in principle, the matrix should turn out exactly the same for no matter which amount of calibration aberration since it gets normalized anyway. However, as we can see in the plot below, a calibration aberration that is too large does not work as we start going into phase wrapping on the individual segments. This is hard to determine when applying a completely random phase map on all segments, so it's best to stick to a small calibration aberration in any case.
Below, the x-axis shows the local aberration modes for which the matrix is created. Here, this is piston, tip and tilt on the 2-Hex telescope with 19 segments total (I use a piston-only matrix for the pytests). The y-axis shows the contrast contribution of each individual mode, here in contrast/nm**2.
The differently colored lines corrsepond to the matrix diagonal when calibrated with different amplitudes of WFE calibration aberration. The lines for 0.1 nm, 1 nm and 10 nm (applied on a single segment when building the PASTIS matrix) are overlapping.
Screenshot 2023-03-14 at 15 25 03

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,7 +8,7 @@ dependencies:
   - configparser
   - flake8
   - jupyter
-  - matplotlib
+  - matplotlib<=3.5
   - numpy
   - openpyxl
   - pandas
--- a/pastis/config_pastis.ini
+++ b/pastis/config_pastis.ini
@@ -193,7 +193,7 @@ lyot_stop_ratio = 0.95
 
 [HexRingTelescope]
 ; aberration for matrix calculation, in NANOMETERS
-calibration_aberration = 100000.
+calibration_aberration = 1.
 ; log10 limits of PASTIS validity in nm WFE
 valid_range_lower = -4
 valid_range_upper = 4
--- a/pastis/tests/data/pastis_matrices/Hex2_efield_coron_imag.fits
+++ b/pastis/tests/data/pastis_matrices/Hex2_efield_coron_imag.fits
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:39bbf77176043015e8c82fd3f9b65a9c572c0f341a09d7bc5c4d384488ab879c
+size 2013120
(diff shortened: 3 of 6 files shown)


ivalaginja:
Hmmm the tests passed locally...

ivalaginja:
Well, at least I have come to a point where the actual test is failing and it's not tripping over some random dependency. It is passing locally though, so maybe I screwed up in uploading the new test files.

ivalaginja:
It was a matter of error tolerances. I had set them very stringently to 1e-24 since of course the matrix I calculate on my machine will be *exactly* equal to another I calculate on my machine as well. If computed on a different machine/OS though, as is the case with Ubuntu used in my Actions workflow, there will be numerical noise. So, I relaxed the errors a bit. If I find another OS requiring another slight relaxation this should be ok.

Thanks to @clementperrot for helping me figure this out.

Comparison of matrices calculated on different machines with different OS:
Screenshot 2023-03-15 at 15 11 55

PR #166, opened 2023-07-05
remove extra normalization factor

### LINKED ISSUE #165: Value of Intensity after fpm is divided twice by the normalization factor.
The intensity saved in  key:‘after_fpm’. is normalized twice. Once in [L1144](https://github.com/spacetelescope/PASTIS/blob/755d218cdf2aa0bf50e7b62500feef144f29c7f2/pastis/simulators/generic_segmented_telescopes.py#L1144) and again in [L1219](https://github.com/spacetelescope/PASTIS/blob/755d218cdf2aa0bf50e7b62500feef144f29c7f2/pastis/simulators/generic_segmented_telescopes.py#L1219)

asahooexo:
The values for intensity obtained just after FPM for a generic-segmented simulator has been divided twice by the maximum intensity just before FPM. This creates flux mis-match (at locations not blocked by coronagraph) between before and after FPM images. 
Fixes #165.

### DIFF
--- a/pastis/simulators/generic_segmented_telescopes.py
+++ b/pastis/simulators/generic_segmented_telescopes.py
@@ -1186,7 +1186,7 @@ class SegmentedAPLC(SegmentedTelescope):
             plt.title('Before FPM')
 
             plt.subplot(3, 4, 9)
-            hcipy.imshow_field(int_after_fpm / wf_before_fpm.intensity.max(), cmap='inferno')
+            hcipy.imshow_field(int_after_fpm, cmap='inferno')
             plt.title('After FPM')
 
             plt.subplot(3, 4, 10)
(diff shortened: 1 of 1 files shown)


ivalaginja (approved):
This looks good to be, thanks for spotting it. Did you double-check that the flux mismatch is gone with this fix?

asahooexo:
@ivalaginja Yes, the fluxes matched to a precision of 13digit after the decimal. 
Here plotting the result before and after the fix. Note that, the intermediate intensity returned  for 'after_fpm' is in logarithm units where as for 'before_fpm' it is in ADU units. I converted 'after_fpm' to ADU units for these plots. 


Before Fix:
Screen Shot 2023-07-11 at 12 53 02 PM

After Fix:
Screen Shot 2023-07-11 at 11 42 34 AM

